PR #22, opened 2015-01-30
Skipping gtf_to_genepred & gene_pred_check, getting version & directory path for eg!

ckongEbi:
To enable eg! FTP pipeline to use the e! GTF dumper module.


magaliruffier:
that looks good, thanks for the change
just wondering why you declared the $eg variable outside of the method?

nerdstrike:
Sneaky package global.... functional but not best practice.

muffato:
Indeed, $self->param() is the recommended way of sharing variables between sections of a Runable

ckongEbi:
Sure i'll replace that with use of self->param

PR #24, opened 2015-04-14
Use auto_increment for the 'attrib_id' column in the 'master_attrib' tab...

ens-lgil:
...le


magaliruffier:
makes sense, I think all other master tables have auto-increment on the primary key

nerdstrike:
I can't see any inserts for this table, so the change probably doesn't need any further changes.

PR #29, opened 2015-09-24
Update compute_closure.pl to allow cross aspect relationships

markmcdowall:
This change moves the requirement for the 2 terms to be part of the same ontology rather than the same namespace/aspect. The means that it is able to capture closures where a term in "molecular_function" is "part_of" a "biological process". See the GO terms GO:0061727, GO:0051596, GO:0019243 and GO:0019172


avullo:
A move forward, that's ok for me, thanks

nerdstrike:
I believe this is safe... hard to foresee the full impact.

PR #32, opened 2015-10-21
Regular expression changes to cater for filename consistency

ckongEbi:
Regular expression changes to cater for filename consistency when turning on per_chromosome flag


magaliruffier:
thanks for the change, sorry I didn't pick up on it earlier

PR #33, opened 2015-10-29
Inter aspect closures to map over is_a and part_of

markmcdowall:
Modified the inter ontology import for retrieving parents of a child term to restrict it to is_a and part_of. In cases where there are more relationships that just the basic 2 this causes querying issues in the marts. For example querying with GO:0008233 would get features annotated with GO:0007866 as well


magaliruffier:
that sounds fine to me

PR #36, opened 2015-12-18
Removed the direct import of a specific ONTO-PERL

markmcdowall:
Term accessions can now include characters other than those represented by the \w regexp. The developer is modifying the ONTO-PERL code and will submit those soon.

This change requires that there is ONTO-PERL in your PERL5LIB. The latest version (1.45) works for me on 5.14.2

### DIFF
--- a/scripts/ontology/scripts/load_OBO_file.pl
+++ b/scripts/ontology/scripts/load_OBO_file.pl
@@ -17,8 +17,8 @@
 use strict;
 use warnings;
 #TODO Do we need this?
-use FindBin;
-use lib "$FindBin::Bin/../../../../ONTO-PERL-1.31/lib";
+#use FindBin;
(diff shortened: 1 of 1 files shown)


nerdstrike:
I approve of removing the hard-coded library version.

Pulling this should be down to @thomasmaurel or @amonida since it affects production procedures.

lairdm:
My comment would be, do we need permanently commented out lines of code in the repository rather than deleting them, version control takes care of "looking back" at previously included libraries should problems arise and debugging necessary.

thomasmaurel:
Looks good to me. Our documentation already specify that we need to have ONTO-PERL in our PATH.
I agree with @lairdm, could you please remove these two lines instead of commenting them out?

Thanks,
Thomas

andrewyatz:
@markmcdowall can you remove rather than comment out please?

magaliruffier:
closing this one as duplicate

PR #37, opened 2016-01-22
Update compute_closure.pl

markmcdowall:
After having a few issues with loading it was identified that there was an error in the closure loading script wrt the handling of links between aspects. I have modified the code in the between aspect and within aspect code to be more respectful and not cross the line.

Quick test load of the db had resulted in the correct output.


magaliruffier:
looks logically sound, I'll leave the rest to production

magaliruffier:
can someone from production ( @thomasmaurel @amonida @ckongEbi ) validate or close this PR?
btw, should occurs_in be part of the default relations?

thomasmaurel:
Looks good to me.

PR #38, opened 2016-02-17
Patch 5 - Removed the lines

markmcdowall:
Updated the patch to just remove the code completely


magaliruffier:
I think this needs to be pushed against master ;-)

PR #39, opened 2016-02-18
Update load_OBO_file.pl

markmcdowall:
Remove the ONTO-PERL specific version requirement


andrewyatz:
Looks better to me

magaliruffier:
thanks

PR #40, opened 2016-03-31
store Ontology term synonym xref separately from name

sarahhunt:
Allows searching by synonym via the API.
Requires https://github.com/Ensembl/ensembl/pull/127

### DIFF
--- a/scripts/ontology/scripts/load_OBO_file.pl
+++ b/scripts/ontology/scripts/load_OBO_file.pl
@@ -184,7 +184,7 @@ sub write_term {
 
   my $statement = "INSERT IGNORE INTO term (ontology_id, subsets, accession, name, definition, is_root, is_obsolete) VALUES (?,?,?,?,?,?,?)";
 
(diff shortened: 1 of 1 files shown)


magaliruffier:
Ok, let's see if I understand this correctly
Currently, the OBO file has lines like
synonym: "mutation affecting regulatory region" RELATED []
this gets stored as type=RELATED and name=mutation affecting regulatory region []
Your change would mean the [] at the end of name is stored separately, or rather, whatever information is in there is stored in the dbxref field
Is that correct?

sarahhunt:
That's right. It means exact matching on synonym will be possible.

magaliruffier:
and we remove the trailing quotes
makes sense, I'm happy with that change

magaliruffier:
The change looks fine, can someone from production please accept the PR if you have no objections? @thomasmaurel @amonida @ckongEbi @danstaines

thomasmaurel:
Looks good to me

PR #41, opened 2016-05-05
Metadata flatfile

lairdm:
Create karyotype files for species, if they exist, but not yet wired in to the dumper pipelines


danstaines:
Best as a TSV file as discussed on Slack

PR #42, opened 2016-05-17
store version for ontology

sarahhunt:
We need to be able to track the versions of the ontologies used in each release. This updates the import script. 
Depends on https://github.com/Ensembl/ensembl/pull/132


magaliruffier:
the schema changes have been merged into core, this change looks fine to me

PR #46, opened 2016-10-05
Version update 86->87 and release 87 patches for production

premanand17:


thomasmaurel:
All good, thanks prem.

PR #52, opened 2017-02-13
Updates to datafiles policy after EBI move

rishidev:
Data files should be group-writable and the group set to "ensemblftp" (hence this is the new default for group checks).  Actual ownership should be irrelevant as long as these permissions and group are maintained.


thomasmaurel:
Looks good to me thanks.

PR #53, opened 2017-02-23
Do not skip first mapping when making chain file

james-monkeyshines:
cf https://github.com/Ensembl/ensembl/pull/184


thomasmaurel:
Thanks for that James.

PR #58, opened 2017-05-08
Revert the escape of some HTML characters

ens-lgil:


thomasmaurel:
Travis error was my fault. All good.

PR #64, opened 2017-06-05
Towards and ensembl-io based RDF dump pipeline. Include type information for BulkFetcher derived 'features'.

avullo:
This is necessary in order to be able to migrate the RDF dump pipeline to using the ensembl-io framework. The ensembl-io based RDF writer needs to be able to access type information of the different 'features' so that it can format and print the appropriate record.


danstaines:
I'll merge this, but you need to come and talk to me about your planned pipeline changes. I'm working on a new search dump pipeline for generating files for the FTP site that will reuse bulk JSON dumped by a master dump pipeline (rather than hitting the database multiple times for the same data) for multiple formats and I want to make sure our work is going to not clash.

PR #65, opened 2017-06-19
Feature/rdf void

avullo:
This is introducing a suite of python modules enabling the generation of a project (i.e. Ensembl or EnsemblGenomes) VOID RDF file which describes the RDF dataset dumped with each release.


avullo:
I'm closing this now as I have introduced other necessary changes to make the whole process more robust, user friendly and less error prone.

PR #68, opened 2017-06-30
RDF pipeline based on ensembl-io

avullo:
Changes have been tested by running the pipeline on the farm with an ensembl species and an eg one from one of the protist collections. Output is comparable with the existing dumps on their respecitive FTP sites.
In terms of differences, apart a slight change in formatting and the order of the triples, I've discovered an inconsistency in the current dumps: triples describing external entities are dumped in the main ttl file, not the xref one. The new pipeline corrects the situation.

I run the tests by invoking directly the RDF pipeline, not the general one (CoreDump).

In order to start using it, it is of course necessary to clone ensembl-io, but then switch to the experimental/unified_design branch.


amonida:
Hi Alessandro, I get a conflict message on your pull request. Can you please have a look? Thanks

avullo:
Thanks Amonida, this is now resolved.

PR #70, opened 2017-07-10
following advice from DS

dbolser-ebi:
Following a suggestion from Dan S to fix an issue with genome loader.


danstaines:
OK, defaults to pepstats without path if not explicitly set elsewhere.

andrewyatz:
Bump?

PR #71, opened 2017-07-28
Updated to remove status from sql query as the column was dropped in …

premanand17:
…e90 from gene table


thomasmaurel:
All good, thanks!

PR #72, opened 2017-08-03
Exclude anything that does not match __\_%

andrewyatz:
We have had issues with linkout not finding some identifiers. So instead we filter out anything that does not look like a normal RefSeq identifier. We also removed the GROUP BY because it was not useful (a check on Ensembl human 89 showed no duplicates between primary accession and stable id).

### DIFF
--- a/scripts/link_out/generate_LinkOut.pl
+++ b/scripts/link_out/generate_LinkOut.pl
@@ -118,7 +118,7 @@ foreach my $host_line (@hosts) {
     $entrez_db = "Nucleotide";
 
     my $sth = $db->prepare(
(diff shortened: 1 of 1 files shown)


thomasmaurel:
@andrewyatz: I just did a re-run of the script with your fix: /gpfs/nobackup/ensembl/maurel/ncbi_linkout_dumps_89_fix but I still get the MICE query in the ressource16 file:

`<Link>
  <LinkId>995870</LinkId>
  <ProviderId>7853</ProviderId>
  <ObjectSelector>
    <Database>Nucleotide</Database>
    <ObjectList>
      <Query>MICE</Query>
    </ObjectList>
  </ObjectSelector>
  <ObjectUrl>
    <Base>&base.url;</Base>
    <Rule>ENST00000505039</Rule>
  </ObjectUrl>
 </Link>
 <Link>
  <LinkId>995871</LinkId>
  <ProviderId>7853</ProviderId>
  <ObjectSelector>
    <Database>Nucleotide</Database>
    <ObjectList>
      <Query>MICE</Query>
    </ObjectList>
  </ObjectSelector>
  <ObjectUrl>
    <Base>&base.url;</Base>
    <Rule>ENST00000510438</Rule>
  </ObjectUrl>
 </Link>`

magaliruffier:
the data has been fixed for human for release 90
Unless we still want to update the dumps for release 89, we don't need this fix any more

thomasmaurel:
Ok thanks @magaliruffier. Since we are going to release in a few days now, I guess it's a bit too late for 89.

PR #74, opened 2017-08-17
[ENSCORESW-2332]. Changed species summary level to be the same as tha…

avullo:
Correct a few issues after EBI-SPOT comments.


thomasmaurel:
Thanks Alessandro!!

PR #76, opened 2017-09-25
Split functionality of SpeciesFactory module into two modules, ...

james-monkeyshines:
...one for flowing db information, one for flowing species information. This is designed to make working with collection dbs effective and efficient (but might also be useful in other scenarios).

The DbFactory flows a "species_list" parameter, and the first element of that list as a "species" parameter. I'm not convinced that this is a good idea, but the idea is for downstream modules which rely on a species parameter to work without modification. For example, the function in the Base.pm which gets a core db adaptor requires a "species" parameter; modules like DatabaseDumper use this to get a connection to the database (on which you can then run non-species-specific operations).

Making the SpeciesFactory not backwards-compatible was a deliberate decision. One could leave the existing functionality in that module, and run it in parallel with DbFactory; but I think that'd lead to more confusion in the long-run, and I think it's better to break things and force appropriate changes to the pipelines. (It now occurs to me that maybe a less destructive approach would be to give the new SpeciesFactory module a different name, to allow for a more manageable transition...)

I've also created a PipeConfig file for demonstrating/testing.

Please let me know what you think...

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbFactory.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbFactory.pm
@@ -1,0 +1,324 @@
+
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2017] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


danstaines:
OK, so here's my thoughts
- we're not really in a position to modify all our existing pipelines in one fell swoop (its also used by others...), so I agree with your suggestion of giving the new factory a different name e.g. DbAwareSpeciesFactory.
- if you keep the old SpeciesFactory, it might make sense to factor out the common code for options processing etc into a base package of some kind?
I'll close this PR and if you can submit a new PR with my suggestions, that'd be great.

PR #77, opened 2017-09-29
Update to build_datafile_ftp_directory.pl so it works with the regula…

mn1:
…tion API in release/91

### DIFF
--- a/scripts/datafiles/build_datafile_ftp_directory.pl
+++ b/scripts/datafiles/build_datafile_ftp_directory.pl
@@ -175,72 +175,72 @@ sub _get_funcgen_DataFiles {
   my ($self, $dba) = @_;
   
   my $species = $dba->species;
-
-  my $segmentation_file_adaptor = $dba->get_SegmentationFileAdaptor;
-  my $result_set_adaptor        = $dba->get_ResultSetAdaptor;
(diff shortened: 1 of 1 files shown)


thomasmaurel:
Looks good thanks

thomasmaurel:
@mn1: I've tried creating the data files symlinks for 91 but it doesn't work for Funcgen, I get the following error:

>         Would have created directory '/nfs/production/panda/ensembl/production/ensemblftp/release-91/data_files/mus_musculus/GRCm38/core/segmentation_file/090/'
> Can't stat /nfs/production/panda/ensembl/production/ensemblftp/data_files/mus_musculus/GRCm38/core/segmentation_file/090: No such file or directory
>  at /homes/ensprod/work/ensembl-production/scripts/datafiles/build_datafile_ftp_directory.pl line 455.

Any idea why we have a "core" in the path?

mn1:
The "core" is added here:

https://github.com/Ensembl/ensembl/blob/release/90/modules/Bio/EnsEMBL/DataFile.pm#L140

mn1:
Not sure how to fix that though

thomasmaurel:
Hmm, I see this is because you are now using the "core" coord_system. Hmm, not an easy fix

PR #78, opened 2017-10-12
[ENSCORESW-1928]. Use simplified and more generic interface.

avullo:
Please, this MUST be merged since the ensembl-io RDF code relies on the new interface.


thomasmaurel:
Thanks Alessandro.

PR #79, opened 2017-11-01
assign ensembl source for MT genes

magaliruffier:
genes with source INSDC will be labelled as coming from Ensembl


amonida:
Done

PR #80, opened 2017-11-01
The species xref section was lacking the dataset attribute since it w…

avullo:
…as again assigned to the core section.


thomasmaurel:
Looks good thanks.

PR #81, opened 2017-11-06
add xmllint fake binary

magaliruffier:
Missing binary was causing Travis to fail
Local runs would work because the actual xmllint binary is usually on the path
With the added fake binary, Travis now passes successfully


thomasmaurel:
and Travis is happy again. Thanks a lot Magali!!!

PR #82, opened 2017-11-06
Making SpeciesFactory 'DB-aware'.

james-monkeyshines:
Database-level operations have been moved from SpeciesFactory to a new parent module, DBFactory.

SpeciesFactory inherits that code, so functions in exactly the same way as previously. However, if your pipeline needs DB-level operations you can use DBFactory directly, and then flow to DbAwareSpeciesFactory, which has almost* the same behaviour as SpeciesFactory, but does not parse the registry, since DBFactory has done that and passed along the necessary data. (* The 'almost' is because DbAwareSpeciesFactory does not flow compara dbs down pipe #5; that behaviour makes no sense, now that DBFactory flows compara dbs down pipe #5.)

Some of the code has been refactored alongside this development, chiefly: the processing of taxonomic parameters has been simplfied; a couple of simple functions for adding/removing species clarify logic and reduce redundancy; detection of chromosomes is done with an API call rather than SQL.

I've created a toy pipeline to test the flow, and have modified the Protein Features pipeline to work appropriately for collecions: the stuff that only needs to be done once (dumping, analyses, xrefs) is only done once, rather than once per species. In order not to break downstream modules, a "species" parameter is still passed to the db-level modules; this is arbitrarily selected from the list of species. Over time, modules that do db-level operations should be changed to not require a species parameter, using instead the db name that is flowed as an output parameter.

I have left the 'check intentions' functionality in the SpeciesFactory, but I do not believe that it should reside there; it caters for a very specific production use case, and would fit better in a dumping-pipeline-specific module. But I don't think I'm the right person to make that change...

Another change which would be good to make, but which I haven't done for fear of breaking extant pipelines, is the SpeciesFactory #1 flow, which outputs a list of species with a parameter named 'species'. This is potentially confusing, because there's a flow #2 parameter, also named 'species', which has a string value. It'd be better to name the former 'all_species', I reckon.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbAwareSpeciesFactory.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbAwareSpeciesFactory.pm
@@ -1,0 +1,54 @@
+
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2017] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


danstaines:
this all looks pretty good to me. As long as we can use existing pipelines without needing to make lots of changes, I'm happy, and we can migrate away as time goes on.

thomasmaurel:
That looks good to me James. Thanks for your hard work. I've created jira tickets for your suggestions as I think they make sense.

PR #83, opened 2017-12-08
Fix for the problem

mn1:
https://github.com/Ensembl/ensembl-production/pull/77#issuecomment-349023424


thomasmaurel:
All good, thanks

PR #85, opened 2018-01-09
Switching eHive branch to 2.4 for Travis testing, as that's what is u…

lairdm:
…sed by Production when running pipelines. This protects against eHive based breakages, as with happened with a recent change on their master branch that made FindBin barf.


lairdm:
Apparently the change to master was intentional. The fix to eHive that's coming should correct the Travis issue.

PR #86, opened 2018-01-10
Pipeline config for transcriptome annotation

james-monkeyshines:
VectorBase use a pipeline to annotate protein domains on transcriptomes (for those at EBI, this pipeline is described in confluence: https://www.ebi.ac.uk/seqdb/confluence/display/EnsGen/Transcriptome+Domains+Pipeline). All of the modules for this pipeline are already in the ProteinFeatures directory in this repo, so it makes sense to also put the configuration there - although the use case is currently only for VectorBase, the code is generic, and will annotate domains on an arbitrary file of DNA sequences.
A couple of minor changes were required to existing modules, nothing controversial, I hope.


thomasmaurel:
Looks good to me, thanks James

PR #87, opened 2018-01-11
Protein Features pipeline email summary

james-monkeyshines:
Make emailing a summary a configurable option (off by default) rather than requiring a separate config file.


thomasmaurel:
Thanks James. This makes perfect sense.

PR #91, opened 2018-01-18
Remove scalar push

nerdstrike:
Pushing to a scalar is classed as experimental in Perl 5.24 . It was probably an accident in this instance.


thomasmaurel:
Thanks Kieron.

PR #92, opened 2018-01-23
[ENSCORESW-2577] Try to expose the real error which leads to intermit…

nerdstrike:
…tent Travis errors. Merge not required. Just need a Travis run to fail...


nerdstrike:
Success! You can merge this in now. It does unavoidably increase the test suite runtime a bit. We might be able to tune it back down again in future.

thomasmaurel:
Thanks Kieron!!!

PR #93, opened 2018-02-14
[ENSCORESW-2475]. Update specific non coding counts to correctly refl…

avullo:
…ect the total.

See https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2475 for a description of the problem.
The StatsGenerator pipeline updates the total non coding counts in the genome_statistics page, but not the specific non coding ones. This happens every release.
Problem is, the specific counts are updated by another pipeline which runs only if there's a change in the gene set, as we can see from different timestamps of the total/specific counts in the genome_statistics table for human.
As a result, the specific counts no longer reflects the total but web displays these together with the correct total which has been updated.


avullo:
As discussed with Thomas, the pipeline has already run but might be worth rerunning it again to have the correct values for 92.

thomasmaurel:
Thanks Alessandro, look good to me. GB have taken some core from staging so I will run the pipeline when these get handed over.

avullo:
As you please :+1:

avullo:
Good the test suite has intercepted some problems. Looks like it's getting exactly half of the expected quantities.

PR #94, opened 2018-02-14
Hiveified xref pipeline

magaliruffier:


thomasmaurel:
Hi @magaliruffier,
We reviewed this and think that it's look good. We are just a bit concerned about two things:
1) In "ParseSource", https://www.ebi.ac.uk/seqdb/confluence/display/ENSCORE/The+Hive+Xref+pipeline#TheHiveXrefpipeline-ParseSource
You mention something about a "race condition", do you have more details about that? Does the jobs die of unknown? We think multiple jobs should be able to read a file at the same time fine. This sounds quite bad as it means we have to keep an eye on the pipeline and figure out if the failure is ok or not.
If the job dies of unknown, it could be because the module loose connection with the hive or core db. We've seen this multiple times with our pipelines, the only solution was to add a `disconnect_if_idle` on both the hive db and core db when you are done using them.
2) In "ParseSource" again, you mention something about using "disconnect_when_idle" which is not great as we have seen in the past strange behaviour when using this, particularly job dying of unknown randomly. Could `disconnect_if_idle` be use on the core db before starting the parsing? That solve a similar issue I had with the xref projection pipeline was used.

magaliruffier:
Regarding the race condition, while I see it as a distinct possibility, I have not seen any jobs failing in that manner once the pipeline was stable. In particular, it did not happen for this release when we were running for 96 species in parallel.
It is also worth noting that we are still planning to run the pipeline for release 93 ourselves, to ensure that it there are no known issues before we hand it over.

Regarding your second point, disconnect_if_idle is used consistenly across all the runnables. The only exception is for the ParseSource stage, where there is currently no disconnect system in place at all. This is because the Database object created for the xref database does not inherit from DBConnection, so those options are not available. This can cause a large number of connections to stay open and has been mitigated by keeping the hive_capacity down to 300. We have a ticket to replace this code with a correct DBConnection object, which will allow us to use disconnect_if_idle (I initially wrote disconnect_when_idle, which was a typo) on the xref database as well as the core database and increase the number of concurrent jobs. We aim to have this change in place before handing over the pipeline.

thomasmaurel:
That sounds perfect, thanks for the explanation.

PR #96, opened 2018-02-15
[ENSCORESW-2475]. Override run method as the transcript and noncoding…

avullo:
… parts are not the object of SnpCount processing. This is also preventing the SnpCount (or any other) pipeline to update noncoding global counts while leaving intact the specific ones.

This is in line with the latest findings as it does seem the problem comes from running GeneCount at one stage in the release and later SnpCount at variation handover. For some, yet unknown, reason, SnpCount recomputes global noncoding counts but at that stage that figure is not reflecting the total that comes by summing the individual (i.e. *_[slm]) values.

The changes presented here separates the two concerns entirely: SnpCount does not update and gene nonglobal count.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Production/GeneCount.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Production/GeneCount.pm
@@ -26,6 +26,92 @@ use Bio::EnsEMBL::Utils::Scalar qw(assert_ref assert_integer wrap_array);
 
 use base qw/Bio::EnsEMBL::Production::Pipeline::Production::StatsGenerator/;
 
+sub run {
+  my ($self) = @_;
+  my $species    = $self->param('species');
+
+  $self->dbc()->disconnect_if_idle() if defined $self->dbc();
+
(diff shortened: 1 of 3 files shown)


thomasmaurel:
Thanks that look good and I trust you ;-)

avullo:
Thanks Thomas, are you planning to rerun the pipeline soon? Might be worth testing it on a copy, or whatever you find it appropriate.

thomasmaurel:
@avullo: The Core or variation part? I can re-run the core stats on everything if that can help. For Variation stats, I would prefer to wait for the variation handover

avullo:
Yes, sure. The core part, in order to check whether the figures match with the already performed run.

thomasmaurel:
Sure, I will run it today.

thomasmaurel:
All done now @avullo. Could you please have a quick look at the core dbs?

avullo:
Thanks @thomasmaurel, I've checked human on staging and all non coding counts add up correctly. It is now a question of verifying the SnpCount pipeline at variation handover.

thomasmaurel:
Thanks @avullo

PR #98, opened 2018-02-19
Replace Bioperl 1.2.3 with 1.6.924 in test environment

nerdstrike:
In anticipation of standardising on 1.6.924, production can now test on the newest safe version of BioPerl. 1.7.2 currently blocked by a codon issue https://github.com/bioperl/bioperl-live/issues/266


thomasmaurel:
Cool, thanks for that.

PR #99, opened 2018-02-23
Automatically generate schema diagrams with GraphViz

muffato:
Here is the code I've used to generate the schema diagrams for Compara:  http://www.ensembl.org/info/docs/api/compara/compara_schema.html
We don't have to fiddle with MySQL Workbench any more !


muffato:
Another important change that comes with this: foreign keys are now parsed from the .sql file

ens-lgil:
Indeed adding the foreign keys in the table description is a big change ... but it makes sense to have them here.

scripts/sql2html.pl
```diff
@@ -49,6 +52,7 @@
     'format_headers=i' => \$format_headers,
     'sort_headers=i'   => \$sort_headers,
     'sort_tables=i'    => \$sort_tables,
+    'out_diagram_dir=s'=> \$out_diagram_dir,
```
ens-lgil: Could you add a description of this new option in the "usage()" method, please ?

muffato: Good point. I've added it

ens-lgil:
Thank you @muffato !
Do you know if the other teams would be interested in this change ?
@magaliruffier and @makostadima, are you happy with this new feature which implies to add/write the foreign keys within the table definitions ?

magaliruffier:
It will probably fail for a while for us until we get a better grasp of all our foreign keys, but I'll put a ticket in for us to update this
No objection from Core

makostadima:
I can't see a reason for not including the foreign keys in the table.sql file. It won't work for us either straight away, as these are missing but we can add them next release.

muffato:
Actually, in eHive (where I first implemented this functionality), the foreign keys are not in the main table.sql, but in a separate file. I didn't bring this code along as Compara's tables.sql has both the schema and the foreign keys, but I'll add the option now, since it seems to be useful for other teams.

Just to make sure we're on the same page: The extra option will allow parsing a file such as https://github.com/Ensembl/ensembl-hive/blob/master/sql/foreign_keys.sql, i.e. with `ALTER TABLE ... ADD FOREIGN KEY ... REFERENCES ...` statements

makostadima:
Even better to have the extra option to parse them from another file. Regulation already has a foreign_keys.sql file that requires some update for next release.

ens-lgil:
Good idea @muffato !
We also have the foreign keys stored in a separate file: https://github.com/Ensembl/ensembl-variation/blob/master/sql/foreign_keys.sql

muffato:
Added

muffato:
I've added the functionality

muffato:
Unless more changes are required, can someone in @Ensembl/production merge this ?

thomasmaurel:
Sure

ens-lgil:
@muffato, which version of ensembl-hive do you use to generate the graphs ?
I got this error: 
"Method cluster_2_attributes not defined! at ensembl-production/scripts/sql2html.pl line 517."

muffato:
master, which will soon be 2.5

ens-lgil:
Thank you @muffato , it works!

PR #103, opened 2018-06-29
Changes after WBPS11 run

wbazant:
Do take what you think is good from these changes. I can edit them or make a pull request against the right branch as appropriate.

## Take a large and unncessary query from RNACentralMapping/UniParcMapping

I had a run where this query was hanging - description in the commit.

## skip_download behaves differently
This was to give me a run I wanted:
- I misconfigured UniProt sources: they had priority that wasn't 1 or 2, so they were downloaded, but not scheduled for parsing
- I changed the config and introduced an error in it by accident - an extra } - and started a run with -skip_download
- The pipeline only broke on the malformed config on a -dump_xref job, next day
- I fixed the config, but realised the sources were bad in the source database, so -skip_download was no good
So the change makes the pipeline skip what I wanted it to skip.

I think this behavior is more useful for other cases too. Similarly -reuse_db 1 breaks the run if there's no source db: the run with -reuse_db 1 could just check if a database is present, and only create it if not.
Maybe such -skip_download and -reuse_db could even be defaults: the user can always delete the sources / the db for a clean run.

## more analysis and hive capacity parameters

I saw jobs failing from too many connections, the extra parameters try to make the pipeline more robust without limiting it much.


marcoooo:
@magaliruffier Is this ok that the PR is for feature/wormbase_xrefs  branch

magaliruffier:
Yes, it is. I have created a feature branch to trial some changes to the pipeline that might make it work for Wormbase. Wojtek has been testing those features on WormBase and has suggested some improvements. I will review these changes and if everything works fine, I can merge them in this feature branch then submit a PR to master to integrate those changes in future runs

marcoooo:
@magaliruffier Do you think we can merge it, or do we wait for another PR for master instead ?

magaliruffier:
it is probably fine to merge it into the feature branch, although I have not had a chance to fully test it yet. It should definitely not go into master yet, as we should run the current pipeline for release 94 first

PR #105, opened 2018-07-02
ENSCORESW-2780: schema update to 94

magaliruffier:
Update schema for release 94.
All test databases patched to the latest schema.
This change would preferably be integrated before branching


thomasmaurel:
Thanks @magaliruffier

PR #107, opened 2018-07-09
remove UCSC source

magaliruffier:
Remove UCSC source from xref pipeline
There is currently an issue when mapping UCSC xrefs to Ensembl so it is safer to leave the source out altogether


thomasmaurel:
Thanks for this.

PR #110, opened 2018-08-03
Xrefs: Import the versioning schema from VersioningService

mkszuba:
This is so that the xref pipeline no longer depends on the
VersioningService repository for a single file. See

https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2728 .

The imported file is the latest available version as of 2018-08-02:
VersioningService commit 6377671892cf897b70fb1a6603dc4ee398818a58 ,
sha256sum ee6a7335fa3835895b721e4b6a1926b4d7db04f645cec8eed5dd59443191b834 .

For the time being the original is still present in VersioningService
but its use is deprecated, it will be removed at some point in the near
future.

Note: this file needs a copyright update but either leave it until after
the version in VersioningService has been removed or update it in both
repos.


mkszuba:
OK, moved the sql directory as suggested.

PR #114, opened 2018-08-03
Feature/unified search dumps

marcoooo:
Please do not merge it right now, this is just to keep aware that we have to merge it as soon as possible.


marcoooo:
We will proceed with a PR to another master copy first.

PR #116, opened 2018-08-08
Feature/wormbase xrefs

magaliruffier:
## Requirements

- Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion;
- Review the [contributing guidelines](https://github.com/Ensembl/ensembl/blob/release/90/CONTRIBUTING.md#why-could-my-pull-request-be-rejected) for this repository; remember in particular:
    - if you change the schema you must patch the test databases as well, see [Updating the schema](https://github.com/Ensembl/ensembl/blob/release/90/CONTRIBUTING.md#updating-the-schema)

## Description

These changes make the xref pipeline more division and species aware.

## Use case

The existing pipeline is very vertebrate centric. The changes are more compatible with other divisions and have been tested by WormBase on a large number of species.

## Benefits

The pipeline can be run on any Ensembl division.

## Possible Drawbacks

## Testing

_Have you added/modified unit tests to test the changes?_
The pipeline was run successfully on a large number of vertebrates and a large set of worms

_If so, do the tests pass/fail?_

_Have you run the entire test suite and no regression was detected?_
NA


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_conf.pm
```diff
@@ -41,22 +41,20 @@ sub default_options {
            ## 'job_factory' parameters
            'species'          => [],
            'antispecies'      => [qw/mus_musculus_129s1svimj mus_musculus_aj mus_musculus_akrj mus_musculus_balbcj mus_musculus_c3hhej mus_musculus_c57bl6nj mus_musculus_cbaj mus_musculus_dba2j mus_musculus_fvbnj mus_musculus_lpj mus_musculus_nodshiltj mus_musculus_nzohlltj mus_musculus_pwkphj mus_musculus_wsbeij drosophila_melanogaster caenorhabditis_elegans saccharomyces_cerevisiae/],
-           'division'         => [],
+           'division'         => 'EnsemblVertebrates',
+           'taxon'            => [],
```
james-monkeyshines: This is a bit confusing. There is a 'taxon' parameter in the SpeciesFactory, so the placing of this variable here implies that's what this is for. But this variable isn't passed to that module, it's used to look for an appropriately named mapping file in ensembl/misc-scripts/xref_mapping/XrefMapper, e.g. "wormbase.pm". And as far as I can make out, the code requires a string value, so I don't know why the default value is an empty arrayref.
So naming the parameter 'taxon' doesn't seem right; I would suggest 'mapping_group'.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf;
+
+use Bio::EnsEMBL::ApiVersion qw/software_version/;
+
+
+sub default_options {
+    my ($self) = @_;
+
+    return {
+           %{ $self->SUPER::default_options() },
```
james-monkeyshines: We're inheriting here from the generic conf, but a lot of the parameters below are simply copied from that file, e.g. email, release, sql_dir. It'd be nicer if we only defined the parameters that are wormbase-specific in this file. (I have a pending change in another branch for the sql_dir parameter, for example, and I don't want to inadvertently have it overwritten here.)

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::Production::Pipeline::PipeConfig::Xref_update_metazoa_conf;
```
james-monkeyshines: Wouldn't this be more appropriately named Xref_update_wormbase_conf?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+
+    return {
+           %{ $self->SUPER::default_options() },
+           'email'            => $self->o('ENV', 'USER').'@ebi.ac.uk',
+           'release'          => software_version(),
+           'sql_dir'          => $self->o('ENV', 'HOME')."/work/lib/ensembl/misc-scripts/xref_mapping",
+           'pipeline_name'    => 'metazoa_xref_update_'.$self->o('release'),
+
+           ## 'job_factory' parameters
+           'species'          => ['caenorhabditis_elegans'],
```
james-monkeyshines: I'm generally opposed to having defaults for the species when using the SpeciesFactory module, because results can be surprising. For example, say you wanted to run this pipeline for all worms, you might set division to 'metazoa'. But because of the order in which parameters are processed, the division parameter will end up getting ignored, and you'd end up just processing C. elegans.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+           'antispecies'      => [],
+           'division'         => [],
+           'taxon'            => 'wormbase',
+           'run_all'          => 0,
+           'force'            => ['caenorhabditis_elegans'],
```
james-monkeyshines: The SpeciesFactory module, to which this is passed, does not have a force parameter. I've got a pending fix that removes this option from the config.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+           'check_intentions' => 0,
+
+           ## Parameters for source download
+           'config_file'      => $self->o('ENV', 'HOME')."/work/lib/ensembl-production/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/xref_sources_metazoa.json",
+           'ftp_release'      => 'WS263',
+           'source_dir'       => $self->o('ENV', 'HOME')."/work/lib/VersioningService/sql",
+           'source_url'       => 'mysql://ensadmin:ensembl@mysql-ens-core-prod-1:4524/mr6_metazoa_source_db',
```
james-monkeyshines: I do not like having default server details in conf files at all, particularly ones with passwords (even if the password is almost as easy to guess as '1234'). If different people/groups are running the pipeline, it's too easy to forget to overwrite with the server where you _really_ want it.

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/RNAcentralMapping.pm
```diff
@@ -31,14 +32,24 @@ sub run {
   my $xref_url     = $self->param_required('xref_url');
   my $species      = $self->param_required('species');
   my $base_path    = $self->param_required('base_path');
-  my $source_url   = $self->param_required('source_url');
+  my $db_url       = $self->param_required('source_url');
```
james-monkeyshines: This is confusing. Why have the local variable named differently to the hive variable?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
```diff
@@ -0,0 +1,64 @@
...
+           'antispecies'      => [],
+           'division'         => [],
+           'taxon'            => 'wormbase',
+           'run_all'          => 0,
+           'force'            => ['caenorhabditis_elegans'],
+           'check_intentions' => 0,
+
+           ## Parameters for source download
+           'config_file'      => $self->o('ENV', 'HOME')."/work/lib/ensembl-production/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/xref_sources_metazoa.json",
+           'ftp_release'      => 'WS263',
```
james-monkeyshines: It seems dangerous to have a default value for ftp_release; presumably that's going to change frequently, so we'd need to update the conf file every release to keep in sync? It can be overwritten at instantiation, of course, but if you forget to do that, then you'll load old data by mistake, and maybe not even notice.

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
```diff
@@ -256,6 +265,25 @@ sub get_taxon_id {
   return $species_id;
 }
 
+sub get_division {
```
james-monkeyshines: I think this would be better named as get_division_id, because that's what it's returning; as it is, it's easy to confuse with the meta_container->get_division method (which this method itself uses...)

magaliruffier:
@wbazant did you use the Xref_update_metazoa_conf.pm file at all or just a local copy of Xref_update_conf.pm?
While this helped test it on non-vertebrate, it could be that it is not required here
I'll check the other sections

wbazant:
@magaliruffier I had no use for Xref_update_metazoa_conf.pm. The commit f4e41eb428, the one adding the -taxon option, is also of no help to us.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_conf.pm
wbazant: @james-monkeyshines I haven't used the option -taxon. I suggest the pipeline not to have the option at all.

I understand providing the option tried to address was to let us specify -taxon wormbase, have the pipeline recognise the mapper XrefMapper::wormbase. It's not helpful to have this supplied pipeline wide, because we have two custom mappers: "XrefMapper::wormbase" and "XrefMapper::parasite". We mostly have the two parser for the sake of nice gene descriptions, which are different for the two groups.

In a single run I would like to use "XrefMapper::wormbase" for WormBase core species gene descriptions, and "XrefMapper::parasite" for gene descriptions of other species. I don't think we will ever have a run that needs a single mapper, specified through the -taxon option

For now I make a separate mapper for each species by duplicating the mapper I want, and put all of these on PERL5LIB. Thus I can have one pipeline run instead of two, give each species nice descriptions, and I'm happy enough. I don't know what a better solution for this would be - but e.g. if the xref pipeline didn't also create gene descriptions, we wouldn't need two mappers.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
wbazant: The option ftp_release was provided so that the pipeline can s/WBRELEASE/WB265/ to once configured paths. I have not used it, I would rather not rely it:
  + I still need to change WormBase sources in `xref_sources.json` whenever WormBase has new species
  + I still need to edit the `xref_config.ini` to make the old code understand the species: we can have many species per taxon, and WormBase core species have a species-specific source

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_metazoa_conf.pm
wbazant: I have no need for this file, and will not need a config checked into the repository as long as there's a good Ensembl config there that I can filter and extend with a small program I wrote.

magaliruffier:
Thanks for the feedback, I will remove the Xref_update_metazoa_conf.pm file as well as the ftp_dir and taxon options.
I will need to test again I have not missed anything, so might be a day or so

magaliruffier:
I am creating a separate branch which should be cleaner

PR #117, opened 2018-08-08
Added generation of a new containing all the species without ortholog…

thomasmaurel:
…s, more details on JIRA ticket: ENSPROD-1482. Pipeline will run as usual generating ortholog files. Once this is done, it will gather a list of all the species for a given division, gather the list of species that we have projected to, compare both list and store species without orthologs into a new file called genome_no_orthologs.txt


thomasmaurel:
Thanks for checking. That's a good point @james-monkeyshines, I would blame the person who created the jira ticket in the first place for the naming ;-). I am a bit reluctant to change the file name since it's now in the GOA pipeline but I've added a description in the module and specified that the file contains species without orthology-based projection. I've also updated the documentation to reflect this: https://www.ebi.ac.uk/seqdb/confluence/display/GTI/DumpOrthologs+Pipeline

PR #119, opened 2018-08-16
Feature/xref update

magaliruffier:
This adds support for non-vertebrate divisions and includes the changes by Wojtek hopefully without any of the complications


modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
```diff
@@ -83,17 +84,17 @@ sub download_file {
...
-    open OUT, ">$file_path" or die "Couldn't open file $file_path $!";
-    my $http = HTTP::Tiny->new();
-    my $response = $http->get($uri->as_string());
-    print OUT $response->{content};
-    close OUT;
+    unless ($skip_download_if_file_present and -f $file_path) {
+      open OUT, ">$file_path" or die "Couldn't open file $file_path $!";
+      my $http = HTTP::Tiny->new();
+      my $response = $http->get($uri->as_string());
+      print OUT $response->{content};
```
lairdm: I realize this is just wrapping old, existing functionality in an unless block, but sucking the file in the memory then spitting it out like that just feels... dirty. HTTP::Tiny has a mirror() routine.

https://metacpan.org/pod/HTTP::Tiny#mirror

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Checksum.pm
```diff
@@ -33,6 +33,12 @@ sub run {
   $self->dbc()->disconnect_if_idle() if defined $self->dbc();
   my ($user, $pass, $host, $port, $source_db) = $self->parse_url($db_url);
   my $dbi = $self->get_dbi($host, $port, $user, $pass, $source_db);
+  if ($skip_download) {
+    my $sth = $dbi->prepare("select 1 from checksum_xref limit 1");
+    $sth->execute;
+    my $table_nonempty = $sth->fetchrow_array;
```
lairdm: Perhaps
return $sth->rows > 0;
or even shortcut with
return $sth->rows
since not zero is true, just to clean it up a bit.

lairdm:
Those two comments are just suggestions for improvement, not critical, feel free to just say no and I'm fine with things as they are. The only other thing that comes to mind is some sanity checking in parse_url(), what happens if the url in the config is malformated. Does it catch the error gracefully. Otherwise, seems fine.

magaliruffier:
Please add the suggestions into JIRA as future improvements
@james-monkeyshines can this be merged into Production as it is?

PR #123, opened 2018-08-31
We should not run xrefs on mus_musculus_casteij, it should be in the …

thomasmaurel:
…antispecies list with the other mouse strains


magaliruffier:
sorry, I thought castaneus was still a separate species, but it does not have a dedicated species page, I stand corrected
Does this mean we have now lost the former projection xrefs by running the regular xref pipeline?

thomasmaurel:
@magaliruffier: Yes, I am a afraid. @amonida is restoring the release 93 database at the moment.

magaliruffier:
OK, let me know if I can help with this

thomasmaurel:
Thanks @magaliruffier, all fixed now.

PR #126, opened 2018-09-03
get the base_dir from the parameters instead of using getcwd

amonida:


thomasmaurel:
That make perfect sense. Thanks for that.

PR #130, opened 2018-09-12
ENSCORESW-2335: say RNAcentral also in source database

wbazant:
I had a run fail because my .ini still said RNACentral, and I went to find out which one it is.


magaliruffier:
if you change it here, it will need changing everywhere else in the xref code as well
I know it is mispelt in the original code, we already updated the display_name in the production database a couple of years back. This change affects the logic_name, which is used in a few places but is not really exposed to our external users

PR #151, opened 2018-10-04
Xrefs: fix --reuse_db and --skip_download

wbazant:
I would like to have these options set to 1 as a default in WormBase ParaSite data production. But they're not working correctly for the first run, for minor reasons. I broke --skip_download myself actually with a bad merge.
This fixes the two. @magaliruffier do have a look!


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/Xref_update_conf.pm
```diff
@@ -116,6 +116,7 @@ sub pipeline_analyses {
              -flow_into  => { '2->A' => 'schedule_source',
                               'A->2' => 'schedule_dependent_source'},
              -rc_name    => 'small',
+             -meadow_type => 'LOCAL',
```
magaliruffier: I seem to remember an attempt at removing all LOCAL meadows from production pipelines, is this just a missing merge?

wbazant: I've only found out about recently that local meadows and Hive 2.5 don't go that well. I'll remove the commit!

wbazant: Removed the commit. Thank you for catching this in the review!

PR #168, opened 2018-11-10
Add EG pan taxonomic compara to search dumps

nicklangridge:
Previously only default or divisional compara was supported.
Pan taxonomic compara has been added as a special case - maybe it will go away in future.
Currently orthologues are the ony data dumped from pan compara.


nicklangridge:
Forget this for now, we'll make another PR later that includes more modifications. @marcoooo let us know when you are ready to accept PRs

PR #182, opened 2018-12-05
Increment InterProScan version

james-monkeyshines:


james-monkeyshines:
I needed to make this change in order to run the pipeline for this release; the pipeline gives an error otherwise. So it seemed appropriate to have the release/95 branch reflect the version that was used in that release.

marcoooo:
> I needed to make this change in order to run the pipeline for this release; the pipeline gives an error otherwise. So it seemed appropriate to have the release/95 branch reflect the version that was used in that release.

Ok then!

PR #183, opened 2018-12-05
Updating analysis/hive capactity for the Protein Features pipeline

james-monkeyshines:
To throttle some over-exuberant analyses, and take advantage of increased farm resources for the analyses doing the grunt work.


james-monkeyshines:
https://ensembl-hive.readthedocs.io/en/master/running_pipelines/management.html#hive-capacity-vs-analysis-capacity

PR #205, opened 2019-01-24
Moved cleanup SQLs into a new Base.pm module. Moved cleanup of GOA GO…

thomasmaurel:
… terms before updating the analysis to make sure that we delete previous release data. Kept a cleanup in LoadFile in case the module fails and need to be restarted (re-running this module multiple time would create duplicates). Increased the number of jobs per analysis. Renamed delete_existing parameter to avoid confusion. Updated Cleanup SQLs to only remove GOA data, we should keep interpro2go data

## Requirements

## Description

See ENSPROD-2684 for details.
In Release 95 the GPAD did not cleanup GO terms properly for Drosophila, S.cerevisia and C.elegans. The following changes fix the cleaning up step and improve it. 
Moved cleanup of GOA GO terms before updating the analysis to make sure that we delete previous release data. Kept a cleanup in LoadFile in case the module fails and need to be restarted (re-running this module multiple time would create duplicates). Increased the number of jobs per analysis. Renamed delete_existing parameter to avoid confusion. Updated Cleanup SQLs to only remove GOA data, we should keep interpro2go data

## Use case

In Release 95 the GPAD did not cleanup GO terms properly for Drosophila, S.cerevisia and C.elegans. The following changes fix the cleaning up step and improve it. 

## Benefits

The pipeline will now work as expected but will also only remove GOA GO xrefs from the database and not all the GO xrefs.

## Possible Drawbacks

None. I've tested on 10 species with success.

## Testing

Dependencies
------------


thomasmaurel:
> Looks to me like it'll work fine for single-species core dbs, but not with collection dbs. Not sure if this is relevant for this pipeline? The analysis table is shared by all species in a collection db; for each species you iterate over, you will be deleting the analysis the previous species uses to link GO terms via object_xref.
> To be collection-safe, you need to do the operations on the analysis table once, at the database level, e.g by using DbFactory as in the Protein Features pipeline.

@james-monkeyshines: That's a very good and valid point thanks. Yes, we need to support collections in this pipeline. I think we got away with it in 95/42 because we didn't run the GPAD load for Fungi...
Back to the drawing board then.

thomasmaurel:
screenshot 2019-01-29 at 12 58 27

thomasmaurel:
@james-monkeyshines: I think this is good to go now. You can see how the pipeline now looks like above. I did a test with collections from protists and Fungi, some metazoan, plants and vertebrates.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/Base.pm
```diff
@@ -0,0 +1,75 @@
+=head1 LICENSE
```
james-monkeyshines: If I've understood the new config correctly, this is redundant now? The cleaning up is done by a combination of AnalysisSetup and RemoveOrphans?

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/FindFile.pm
```diff
@@ -0,0 +1,64 @@
...
+=head1 AUTHOR
+
+maurel@ebi.ac.uk
+
+=cut
+package Bio::EnsEMBL::Production::Pipeline::GPAD::FindFile;
+
+use strict;
+use warnings;
+use Data::Dumper;
```
james-monkeyshines: Most of these libs aren't used, only Registry...

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
```diff
@@ -22,7 +22,7 @@ Bio::EnsEMBL::Production::Pipeline::GPAD::LoadFile;
 
 =head1 AUTHOR
 
-ckong@ebi.ac.uk
+ckong@ebi.ac.uk and maurel@ebi.ac.uk
```
james-monkeyshines: Is it not Ensembl policy to not have named authors? I think EG worked differently, and this might be a vestigial tail from those times.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
```diff
@@ -34,96 +34,25 @@ use Bio::EnsEMBL::Analysis;
...
-                      WHERE x.external_db_id=1000
-                      AND c.species_id=?
-                      AND (a.logic_name="goa_import"
-                      OR a.logic_name="interpro2go")'];
-
-      for my $sql (@$sqls) {
-	$self->log()->debug("Executing $sql");
-	$dba->dbc()->sql_helper()->execute_update(-SQL=>$sql);
-      }
+      $self->cleanup_GO($dba);
```
james-monkeyshines: As mentioned above, I don't think this is needed anymore.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
```diff
@@ -45,38 +46,37 @@ sub default_options {
...
+    'gpad_directory' => '',
 
 	## Email Report subject
-    'email_subject'       	   => $self->o('pipeline_name').' GPAD loading pipeline has finished',
+    'email_subject'  => $self->o('pipeline_name').' GPAD loading pipeline has finished',
 
     ## Remove existing GO annotations
     # on '1' by default
-    'delete_existing' => 1,
+    'delete_existing_go' => 1,
```
james-monkeyshines: I think it's clearer to have a single 'delete_existing' parameter; if you delete the analysis, you need to delete the associated GO terms anyway, to prevent orphans. And there's no benefit to keeping the analysis if you've deleted all its GO terms.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
```diff
@@ -139,43 +138,114 @@ sub pipeline_analyses {
...
+                }
     },
     { -logic_name     => 'gpad_file_load',
   	  -module         => 'Bio::EnsEMBL::Production::Pipeline::GPAD::LoadFile',
-      -hive_capacity  => 10,
-      -rc_name 	      => 'default',     
-    },   
-
+      -parameters     => {
+                              delete_existing => $self->o('delete_existing_go'),
```
james-monkeyshines: Might be worth passing in the logic_name here, rather than having 'goa_import' hard-coded in the module?

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/Base.pm
thomasmaurel: Actually, this is still needed. If you re-run the "gpad_file_load" analysis multiple times (e.g: because of a failure) then you will end up with duplicated GOA data... This is to make sure we don't end up in this situation. I could move the deletion back to gpad_file_load module.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/FindFile.pm
thomasmaurel: Thanks, I will clean up.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
thomasmaurel: I don't think it is. I could remove this

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
thomasmaurel: As mentioned above. I will move it back to loadFile module

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
thomasmaurel: Sure, that make sense. I changed it to avoid confusion but what you say make total sense.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
thomasmaurel: That's a good point, will do that.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/Base.pm
thomasmaurel: See latest commit.

modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
thomasmaurel: see latest commit

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
thomasmaurel: see latest commit.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
thomasmaurel: see latest commit

PR #208, opened 2019-01-31
Copy service extend server support

thomasmaurel:

## Requirements

## Description

Extending Copy database module to deal with situation when we don't have file system access or if the database is in InnoDB format. For both situations, the module will assess if we have filesystem access on both source and target server. If we have, the current code will be used, database file copy using rsync. If we don't, then the module will use MySQL dump tp dump the source database into a sql file, create database on targer server and then load the file into the target database. Once the database has been copied, mysqlcheck will be run to make sure all is fine and the database is not corrupted. Some code change to make it work when we have file system access or not.

## Use case

We can now use the service to copy database if we don't have file system access or if the database is in InnoDB format

## Benefits

We can copy database from any servers and we can copy both InnoDB and MyISAM db

## Possible Drawbacks

Dumping the database will be slower than directly copying file system. 
There is a slight risk of people handing over database in InnoDB. these would then end up on staging.

## Testing

- [N ] Have you run the entire test suite and no regression was detected?
- [ NA] TravisCI passed on your branch

I did a lot of testing. Copying using servers with file system access and testing all the parameters
Testing when source/target server doesn't have file system access
Testing copy of InnoDB database

Dependencies
------------

No change for code depending on this code.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
+++ b/modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
@@ -44,6 +44,8 @@ if(!Log::Log4perl->initialized()) {
 sub copy_database {
   my ($source_db_uri,$target_db_uri,$opt_only_tables,$opt_skip_tables,$update,$drop, $verbose) = @_;
 
+  # Path for MySQL dump
+  my $dump_path='/nfs/nobackup/ensembl/ensprod/copy_service/';
   # Get list of tables that we want to copy or skip
   my %only_tables;
   my %skip_tables;
@@ -224,7 +226,7 @@ sub copy_database {
       croak "We don't have file system access on these server so we can't use the --update or --only_tables options";
     }
     else{
-      $copy_failed = copy_mysql_dump($source_dbh,$target_dbh,$source_db,$target_db);
+      $copy_failed = copy_mysql_dump($source_dbh,$target_dbh,$source_db,$target_db,$dump_path);
     }
   }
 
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
```diff
@@ -631,14 +675,51 @@ sub copy_mysql_files {
   return $copy_failed;
 } 
 
+# Dump the database in /nfs/nobackup/ensembl/ensprod/copy_service
+# Create database on target server
+# Load the database to the target server
+sub copy_mysql_dump {
+  my ($source_dbh,$target_dbh,$source_db,$target_db) = @_;
+  my $copy_failed=0;
+  my $file='/nfs/nobackup/ensembl/ensprod/copy_service/'.$source_db->{host}.'_'.$source_db->{dbname}.'_'.time().'.sql';
```
james-monkeyshines: It'd be a hassle to have the path as a parameter, but at the same time, it's rather buried in the code. Perhaps a halfway house would be to define a constant at the top of the file, so that it's a bit more obvious, and easier to change in the future if necessary?

modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
```diff
@@ -631,14 +675,51 @@ sub copy_mysql_files {
...
+  $logger->info("Dumping $source_db->{dbname} from $source_db->{host} to file $file");
+  if (defined $source_db->{pass}){
+    if ( system("mysqldump --host=$source_db->{host} --user=$source_db->{user} --password=$source_db->{pass} --port=$source_db->{port} $source_db->{dbname} > $file") != 0 ) {
```
james-monkeyshines: mysqldump will attempt to lock tables unless you tell it not to; but if you're doing this as the ensro user, you won't have permission to do that. Can use the flag `--skip-lock-tables` to avoid this.

james-monkeyshines: Also, I've had trouble in the past with dumping big tables, because the default packet size is too small. That can be changed with something like `--max_allowed_packet=1024M`.

modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
thomasmaurel: @james-monkeyshines: That make perfect sense. You are right, it's a hassle to add it as a parameter especially as it would only very rarely change. I've added the path at the start of the script.

modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm
thomasmaurel: @james-monkeyshines: That's two very good points. For the first one, the script already do a table lock here: https://github.com/Ensembl/ensembl-production/blob/fb901c60639e52c27a9b4e424045784f50246717/modules/Bio/EnsEMBL/Production/Utils/CopyDatabase.pm#L194. I think this is quite important and most of the servers now allow ensro user to flush and lock. We actually have steps in our SOP to add these privileges to ensro users using the admin user. So no need to lock twice here.
I forgot about the max_allowed_packed and it did bite me a few times... I've now updated the command. Thanks.

thomasmaurel:
@james-monkeyshines: Please hold off merging this, I need to test my recent changes once everything is back up at the EBI.

james-monkeyshines:
@thomasmaurel I don't know why this has only cropped up now, as you say, it's nothing to do with these changes. But the .travis.yml file is using a non-existent branch for the samtools/htslib repo, which I suspect is the root cause of the failing perl module. The default branch is 'develop'.

thomasmaurel:
@james-monkeyshines: Very well spotted, I can push a fix for it.

marcoooo:
> @james-monkeyshines: I have no idea how to fix Travis but this has nothing to do with my change...
> 
> > cc -c  -I/usr/local/BerkeleyDB/include -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -O2   -DVERSION="1.843" -DXS_VERSION="1.843" -fPIC "-I/home/travis/perl5/perlbrew/perls/5.14/lib/5.14.4/x86_64-linux/CORE"  -D_NOT_CORE  -DmDB_Prefix_t=size_t -DmDB_Hash_t=u_int32_t   version.c
> > version.c:30:16: fatal error: db.h: No such file or directory
> > #include <db.h>
> > ^
> > compilation terminated.
> > make: *** [version.o] Error 1
> > FAIL
> > ! Installing DB_File failed. See /home/travis/.cpanm/work/1549365175.15709/build.log for details. Retry with --force to force install it.

Thhe issue comes from the Bio:: DB::HTS dependency, core is looking into it

thomasmaurel:
@james-monkeyshines: My last changes from your suggestions (not related to Travis) are working so I think this is ok to merge now

PR #210, opened 2019-02-01
fix: do not leave temporary dot files around

muffato:
## Description

When doing some tests for the Compara documentation, I found that there was a .dot file around. I believe it shouldn't be there.

## Use case

## Benefits

No leftover .dot file hanging around. Unlikely to go and pollute public-plugins

## Possible Drawbacks

If people want the .dot file they have to go and uncomment the line in the source code

## Testing

- [ no ] Have you added/modified unit tests to test the changes?
- [ no ] Have you run the entire test suite and no regression was detected?
- [ hopefully ] TravisCI passed on your branch

Dependencies
------------

None


muffato:
Thanks !

PR #216, opened 2019-02-08
Dumping DNA and running CoreStats on all the species is costly. We ca…

thomasmaurel:
…n now use the metadata database to know which species have new assembly or new genebuild. Created new module called CheckAssemblyGeneset using species name and release to figure out if the species had a new_assembly or new_genebuild update, at the end flow species name and two flags. Updated CoreStatistic and DumpCore pipelines to use this module. CoreStats will only be rn on species with new assembly or genebuild. Dump DNA will only run on species with update assembly, for the other species, the dump from last release FTP site will be copied.

## Requirements

## Description

Dumping DNA and running CoreStats on all the species is costly. We can now use the metadata database to know which species have new assembly or new genebuild. Created new module called CheckAssemblyGeneset using species name and release to figure out if the species had a new_assembly or new_genebuild update, at the end flow species name and two flags. Updated CoreStatistic and DumpCore pipelines to use this module. CoreStats will only be rn on species with new assembly or genebuild. Dump DNA will only run on species with update assembly, for the other species, the dump from last release FTP site will be copied.

## Use case

To save us computation time. Dumping DNA is very costly in term of ressources for no gain if the assembly hasn't changed. The CoreStats are very time consuming for no gain if geneset/assembly has not changed. As we get more species, this change will reduce our compute time and ressources use.

## Benefits

Compute time.

## Possible Drawbacks

If information is wrong in the metadata there is a risk that an updated species would not be re-dumped or core stats would not be re-run.

## Testing

- [ N] Have you run the entire test suite and no regression was detected?
- [ ?] TravisCI passed on your branch

Dependencies
------------

Dependancy on the metadata database and metadata API. The metadata database need to be in the registry file for the CheckAssemblyGeneset module to work.


thomasmaurel:
screenshot 2019-02-08 at 12 19 35
Updated Core stats

thomasmaurel:
screenshot 2019-02-06 at 16 37 49
Updated DumpCore

thomasmaurel:
screenshot 2019-02-08 at 14 26 40
Updated DumpCore Non Vertebrates

thomasmaurel:
Travis need the Base module which is not on the master branch of the metadata repo at the moment. I made this pull request here: https://github.com/Ensembl/ensembl-metadata/pull/44, once this is merged, restarting Travis should make it pass

muffato:
Hi there,
I'm curious if this really works as intended. I thought that the metadata database was relying on the assembly name and the genebuild start-date, meaning that it is blind to updates such as adding the missing MT, adding some missing scaffolds, or genebuild updates like the Havana merge on human and mouse etc. Can you elaborate ?

thomasmaurel:
@muffato: For the assembly, the metadata database code relies on a set of meta key like assembly_name, assembly_default, assembly_accession,... This new module uses assembly_default to figure out if any changes have been made to the assembly, this should capture human and mouse patches update too. For the genebuild, the metadata code is looking at both genebuild.version and genebuild.last_geneset_update. HAVANA merge will be picked up as the genebuild.last_geneset_update meta key will change. You are right that MT addition or fix in assembly will not be picked up as an assembly change at the moment but genebuild.last_geneset_update meta key will change so we still spot that something has changed. I don't know if there is an automatic way to spot these changes to assemblies? I think a temporary way around this would be to re-dump DNA when assembly or geneset change, this would save us re-dumping all the species but would also make sure we don't miss out these species. This PR is actually restoring a feature that we had in the past except the code was looking at the production database declarations which I believe was less reliable than the metadata code.

muffato:
Thanks @thomasmaurel . All good on the gene side, then :+1: 
To test assembly updates (we need this in Compara), I was going to just check whether the toplevel sequences are the same or not, maybe by keeping a checksum in the database

thomasmaurel:
@muffato: For the assembly, we store the assembly sequences in the metadata database so I believe that I can update the metadata API code to compare the sequences between two releases and spot new MT, updated scaffold or any other assembly fix. I will look into implementing this for 97 and all the script/modules using this.

muffato:
Super, thank you !

thomasmaurel:
@muffato: We found a way to detect updates such as adding the missing MT, adding some missing scaffolds. Basically the metadata database populate a field called "base_count" which correspond to the following SQL: 

> select sum(s.length) from seq_region s 
> join coord_system c using (coord_system_id) 
> join seq_region_attrib sa using (seq_region_id) 
> join attrib_type a using (attrib_type_id) 
> where a.code='toplevel' and species_id=?

This is making a sum of all the seq_region length so from this SQL you can detect any assembly sequence changes even if the assembly meta keys have not changed.
I will push a change implementing these new checks in this PR to make sure that we don't miss the Cow MT update this release. report_genomes.pl will also now report Cow as an updated assembly for 96.

thomasmaurel:
Please wait for this PR to be merged: https://github.com/Ensembl/ensembl-metadata/pull/47
Before merging this.

muffato:
Thank you @thomasmaurel, that's really useful ! I think we can now directly use the output of `report_genomes.pl`

thomasmaurel:
Well I hope so @muffato, we have ironed out a lot of bug this release.

thomasmaurel:
Travis is failing because the metadata PR has not been merged yet. Once this is done, I will restart Travis tests and I expect them to pass.

thomasmaurel:
Travis will work once this new PR is merge to have recent code on master: https://github.com/Ensembl/ensembl-metadata/pull/48
Promise.

modules/t/test-genome-DBs/multi/metadata/organim.txt
```diff
@@ -0,0 +1 @@
+170	9606	0	9606	homo_sapiens	Homo_sapiens	Human	Homo sapiens	NULL	NULL	NULL	NULL
```
james-monkeyshines: Typo in filename, 'organism.txt' not 'organim.txt'.

modules/Bio/EnsEMBL/Production/Pipeline/Common/CheckAssemblyGeneset.pm
```diff
@@ -0,0 +1,91 @@
...
+use warnings;
+use Bio::EnsEMBL::MetaData::DBSQL::MetaDataDBAdaptor;
+use Bio::EnsEMBL::MetaData::Base qw(process_division_names fetch_and_set_release check_assembly_update check_genebuild_update);
+use base qw/Bio::EnsEMBL::Production::Pipeline::Common::Base/;
+
+sub write_output {
+  my ($self) = @_;
+  my $release = $self->param_required('release');
+  my $species = $self->param_required('species');
+  my $run_all = $self->param('run_all');
```
james-monkeyshines: I don't think we should be re-using the parameter name 'run_all'; that is used with Db/SpeciesFactory modules, and has a different meaning here. Suggest changing to 'always_run'.
Also, this parameter value could be tested right at the top of the method; if it's set, we don't need to bother attaching to the metadata db to try to find a previous version.

modules/Bio/EnsEMBL/Production/Pipeline/FASTA/Base.pm
```diff
@@ -33,10 +33,25 @@ sub fasta_path {
 sub old_path {
   my ($self, $species, $directory) = @_;
   my $base = $self->param('ftp_dir');
+  my $eg = $self->param('eg');
```
james-monkeyshines: Introducing this parameter seems unnecessary; we're trying to erode the EG brand, after all. Why not just test if `$self->division` is 'vertebrates', and use that to determine the url path?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
```diff
@@ -130,38 +130,65 @@ sub pipeline_analyses {
...
+      -priority        => 5,
+      -rc_name         => 'normal',
+    },
     {
       -logic_name      => 'GeneRelatedTasks',
       -module          => 'Bio::EnsEMBL::Hive::RunnableDB::Dummy',
       -max_retry_count => 1,
       -flow_into       => {
-                            '1->A' => [
+                            '1->A' => WHEN( '#new_assembly# >= 1 or #new_genebuild# >=1' => [
```
james-monkeyshines: Why not put this one level further up, in the CheckAssemblyGenesetGeneTask's flow to 'GeneRelatedTasks'? That would simplify the logic a bit, only need one conditional.
Same argument applies to CheckAssemblyGenesetChromosomeTasks.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
```diff
@@ -130,38 +130,65 @@ sub pipeline_analyses {
...
       -rc_name         => 'normal',
     },
-
+    { -logic_name  => 'CheckAssemblyGenesetGeneTasks',
+      -module      => 'Bio::EnsEMBL::Production::Pipeline::Common::CheckAssemblyGeneset',
+      -parameters  => {
+          run_all => $self->o('run_all'),
+          release => $self->o('release')
+       },
+      -can_be_empty    => 1,
```
james-monkeyshines: A minor point, but I don't think this line is required.

modules/t/test-genome-DBs/multi/metadata/organim.txt
thomasmaurel: Well spotted, thanks

modules/Bio/EnsEMBL/Production/Pipeline/Common/CheckAssemblyGeneset.pm
thomasmaurel: My idea was to reuse this flag but your are right it might not be appropriate. I've added a new flag called "skip_metadata_check" working as you describe above.

modules/Bio/EnsEMBL/Production/Pipeline/FASTA/Base.pm
thomasmaurel: I was using the `eg` flag because it's used consistently in this pipeline but you are right, not using it will be one less line to update when we get rid of it in the future.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
thomasmaurel: Very good point, these dummy are not really needed anymore. I've cleaned up these analysis.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
thomasmaurel: I've removed this.

PR #217, opened 2019-02-12
We also need to remove dependent xref that could have been linked to …

thomasmaurel:
…object_xref row

## Requirements

## Description

A previous PR updated the GPAD pipeline to properly clean up old GOA data stealing code from the Protein features pipeline. I only took the cleanup of the ontology_xref table but this is not enough as sometime and object_xref can be link to an ontology_xref and dependent_xref (Thanks James for the explanation). This change make sure that we clean up dependent xrefs too. 

## Use case

All the time we run this pipeline, we need to clean up dependent xrefs too or we will endup with FK issues

## Benefits

We won't get FK error anymore after running the GPAD pipeline

## Possible Drawbacks

None really. 

## Testing

- [ N] Have you run the entire test suite and no regression was detected?

Dependencies
------------


thomasmaurel:
Ok, that

PR #228, opened 2019-02-28
Improve travis git branch

thomasmaurel:

## Requirements

## Description
At the moment Travis uses other ensembl repo on master branch. This causes issue when future release changes are committed to master since the release and other branches don't have the associated code. As a result Travis break. With this code, on a release branch Travis will use the repos on this branch. If a PR is created, it will clone using the destination branch name

## Use case
At the moment Travis uses other ensembl repo on master branch. This causes issue when future release changes are committed to master since the release and other branches don't have the associated code. As a result Travis break. With this code, on a release branch Travis will use the repos on this branch. If a PR is created, it will clone using the destination branch name

## Benefits
Travis will seamlessly work on any branch without any need to update the .travis.yml file.

## Possible Drawbacks
If you create a dev branch, travis will run using master branch for the other repos but when creating the PR it will use the destination branch.

## Testing
* [ ]  Have you added/modified unit tests to test the changes?
* [ ]  If so, do the tests pass?
* [ Y] Have you run the entire test suite and no regression was detected?
* [ Y] TravisCI passed on your branch


thomasmaurel:
> I think this should be made simpler, hence more reliable with spendig the 2 mns needed to update from one branch to another. The build is still failing whatsoever.

Hmm I though this was a Draft branch... I agree that the code is too complicated at the moment. I still think that it's worth looking into this instead of manually changing this file forever...

PR #233, opened 2019-03-08
Updated databases names according to renaming

marcoooo:

## Requirements

## Description

Renamed values in pipeline config to fit with new EG database naming scheme

## Use case
N/A

## Benefits

Good names in place

## Possible Drawbacks

## Testing

Dependencies
------------


marcoooo:
> Should the pipeline run on `mysql-ens-sta-3` instead of `mysql-ens-mirror-3`? I am not sure what this pipeline is for.

I used the renamed server which use the same port. let's wait for @james-monkeyshines on Monday to validate.

marcoooo:
> This pipeline would be run to update a compara for the current release; so you wouldn't want to point at mirrors for the source data, it would have to be ens-sta-3.
> It's just that when this pipeline was last updated, eg-s2 was the 'current' server.
> However, this is absolutely not a Production pipeline. It's an EG one which was needed temporarily to bring EG comparas up to spec; but this code is all now in compara pipelines - we need to delete this config, and any modules associated only to it.

Files has been removed.

PR #235, opened 2019-03-14
Extending regex that matches protein accessions to include hyphens

james-monkeyshines:
## Description
Extending regex that matches protein accessions to include hyphens.

## Use case
Needed for successful import of plant species.

## Benefits
Pipeline won't fail...

## Testing
Pipeline with this fix ran successfully.


thomasmaurel:
Travis is failing on release branch because ORM doesn't have a release branch and now contains changes for release 97..

james-monkeyshines:
And here's me, holding off on submitting this for a couple of weeks, until all the travis/release issues were sorted...

thomasmaurel:
That's recent news ;-)

PR #255, opened 2019-04-04
Compara file names changed, update string to match

james-monkeyshines:
## Description
Compara phyloxml files now have ID ranges embedded in the file names. Need to adjust the pattern for matching in the FTP Checker pipeline.


muffato:
Note that this was done manually this time, and for Vertebrates only. I hope we'll have an automated solution in e97: https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2019

FYI I have added a script to our repo + SOP to check that the FTP is growing and that no files have disappeared: https://github.com/Ensembl/ensembl-compara/blob/release/97/scripts/production/compare_ftp_dir.pl
And we'll add another script to check that _every_ file is there: https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2180
There is no harm in you testing this on your side (it's always better to double check and I suppose it's quick anyway). This is just to let you know that we're adding some checks on our side too and that errors should be detected sooner (before we handover the dumps to you)

thomasmaurel:
That's great to hear @muffato. This pipeline actually do very minimal checks for Compara but as you said it doesn't take too much time to run.

PR #267, opened 2019-05-03
Handle empty variation db

thomasmaurel:

## Requirements

## Description

Some variation databases are now empty since the data is coming from a VCF file. Our pipeline cannot deal with these at the moment so we should exclude them. The Variation team have introduced two new meta keys variation.source.vcf = 1/0 and variation.source.database = 1/0. Since we are only interested if the database contain data, we need to check that value of variation.source.database. Only vertebrate division has these meta key so it's important to check the existance of this meta key first. More information here: https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-3161

## Use case

In release 96, Vervet monkey was the first empty variation db. In release 97 the database now has the following meta keys: 
- variation.source.vcf = 1
- variation.source.database = 0
The other vertebrates will have:
- variation.source.vcf = 0
- variation.source.database = 1
The non-vert won't have these meta keys.

## Benefits

Pipelines won't be failing on empty databases anymore.

## Possible Drawbacks

None really. In the future we could implement code to process these dbs if possible/required.

## Testing

- [ N] Have you added/modified unit tests to test the changes?

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/SpeciesFactory.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/SpeciesFactory.pm
@@ -169,7 +169,19 @@ sub has_variation {
   my $dbva =
     Bio::EnsEMBL::Registry->get_DBAdaptor( $species, 'variation' );
 
-  my $has_variation = defined $dbva ? 1 : 0;
+  my $has_variation;
+  if (defined $dbva){
+    my $source_database = $dbva->get_MetaContainer->list_value_by_key('variation.source.database')->[0];
+    if(defined($source_database)){
+      $has_variation = $source_database == 1 ? 1 : 0;
+    }
+    else{
+      $has_variation = 1;
+    }
+  }
+  else{
+    $has_variation = 0;
+  }
   
   $has_variation && $dbva->dbc->disconnect_if_idle();
 
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/VariationStatistics_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/VariationStatistics_conf.pm
@@ -81,6 +81,7 @@ sub pipeline_analyses {
                                         'SnpDensity',
                                       ],
                             'A->1' => ['GenomeStats'],
+                          },
       -rc_name         => 'normal',
     },
 


thomasmaurel:
@james-monkeyshines: I've only updated the SpeciesFactory but please let me know if Dbfactory and DBaware factory needs some update too.

modules/Bio/EnsEMBL/Production/Pipeline/Common/SpeciesFactory.pm
```diff
@@ -169,7 +169,19 @@ sub has_variation {
   my $dbva =
     Bio::EnsEMBL::Registry->get_DBAdaptor( $species, 'variation' );
 
-  my $has_variation = defined $dbva ? 1 : 0;
+  my $has_variation;
```
james-monkeyshines: I'm not convinced that this is the best approach. In the context of the variation stats pipeline this works, because we don't want to do anything with a variation db that has no variants in it. But in other cases, that might not be true - for example, the FTPChecker pipeline uses the variation dataflow in order to determine whether variants should exists for a species. And you could imagine other use cases, e.g. dumping all databases associated with a species, for which this would not dump the variation db, which is properly not what you expect/want.
My suggestion would be to turn the 'VariationTasks' from a dummy module to a proper one, with this source logic inside it, controlling whether anything is flowed.

thomasmaurel: @james-monkeyshines: That is a very good point but the issue is that many pipelines are using Species variation factory including mart pipelines and all these pipelines expect a "complete" variation database. As we have just discovered, they don't generate any dumps for these databases so it should be fine for FTPChecker and it should not check these dbs. For database dumping, we are getting the list of database from the metadata database at the moment. The Metadata pipeline don't process these variation dbs since they don't have any data in it.
I believe it is better to change the species factory than all the different pipelines using it and this would avoid pipelines to die because they can't process these dbs. 
I think Variation should have consulted other teams about the possible side effect of these databases instead of pushing this live.
Alternatively, we could always have a new variation flow with the variation vcf dbs in it so that they are not lost. I think that if variation wants us to process these databases in the future then we would need to implement something new and possibly update our pipelines.

james-monkeyshines: OK, yes, I agree with your argument. Having a separate flow for vcf-only dbs makes sense, but I don't think we need to bother implementing it until we have a use case.

PR #283, opened 2019-06-05
Renamed all the analysis to remove the dump_ prefix. We now have json…

thomasmaurel:
… instead of dump_json. Optimized RDF dumps, now using default memory for most of the jobs, then 15GB, 32GB and finally 64GB. Optimized json and RDF dumps, to try 15GB before 32GB since 32GB and 64GB queues are hard to get on it. Changed how we generate checksums as we can't run checksum in the dump_path anymore since we have many divisions and other dumps in /hps/nobackup2, checksums are now generated on a species basis and only for dumps generated by the pipeline. There are extra code for Blat, ncbi_blast since we have all the species into the main directory and fasta_blast are not in the list of dumps for non-vert.

## Requirements

## Description

Renamed all the analysis to remove the dump_ prefix. We now have json instead of dump_json. Optimized RDF dumps, now using default memory for most of the jobs, then 15GB, 32GB and finally 64GB. Optimized json and RDF dumps, to try 15GB before 32GB since 32GB and 64GB queues are hard to get on it. Changed how we generate checksums as we can't run checksum in the dump_path anymore since we have many divisions and other dumps in /hps/nobackup2, checksums are now generated on a species basis and only for dumps generated by the pipeline. There are extra code for Blat, ncbi_blast since we have all the species into the main directory and fasta_blast are not in the list of dumps for non-vert.
Checksums were generated with a chmod making them non-group writable which is what we want to avoid...
The pipeline layout has changed, please see attached screenshot.

## Use case

We have multiple issues here. For RDF the 64GB queue is really hard to get onto and most species don't need that much memory. For checksums are current system doesn't work anymore in the new /hps/nobackup2 since we now have all the divisions and other type of dumps like compara, variation.
For json we can run many species on less memory. 

## Benefits

The pipeline will be more efficient at generating checksums and it will now work in all situation. RDF dumping will be more efficient too since we will manage memory in a better way.

## Possible Drawbacks

Adding new dumps with subdirectories might requires changes to the ChksumGenerator but I am not sure there are any plans for new dumps at the moment.

## Testing

Tested on a sample of vert and non-vert including collections using 97 data.

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/ChksumGenerator.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/ChksumGenerator.pm
@@ -60,22 +60,24 @@ return;
 }
 
 sub run {
-    my ($self) = @_;
-    my $dirs   = get_dirs($self, $self->param_required('base_path'));
+  my ($self) = @_;
+  my $base_path = $self->param_required('base_path');
+  my $species = $self->param_required('species');
+  my $dumps = $self->param_required('dumps');
+  my $dirs   = get_dirs($self, $base_path, $species, $dumps );
 
-    foreach my $dir (@{$dirs}) {
-      $self->info('Producing CHECKSUM on directory %s', $dir);
+  foreach my $dir (@{$dirs}) {
+    $self->info('Producing CHECKSUM on directory %s', $dir);
 
-      my $contents = get_contents($self, $dir);
+    my $contents = get_contents($self, $dir);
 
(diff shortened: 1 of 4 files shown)


thomasmaurel:
Screenshot 2019-06-05 at 11 03 23

thomasmaurel:
Screenshot 2019-06-05 at 11 03 15

thomasmaurel:
Screenshot 2019-06-05 at 11 04 11

PR #293, opened 2019-06-24
Added new step to sync the external_db data from the ensembl-producti…

thomasmaurel:
…on database to the compara database since this data is required by the pipeline.

## Requirements

## Description

The Gene Tree Highlighting Pipeline require the external_db data from the production database to be in the compara database before running. Since we have stopped cronjobs to sync this table daily, we now need to sync it before running the pipeline.

## Use case

We need this table to be present and populated when running the Gene Tree Highlighting Pipeline.

## Benefits

The pipeline can now run without failures.

## Possible Drawbacks

We need mysql command in order to be able to run this pipeline. It won't be a problem for most people.

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
@@ -45,6 +45,10 @@ sub default_options {
     run_all      => 0,
     antispecies  => [],
     meta_filters => {},
+    production_db => 'ensembl_production',
+    production_host => 'mysql-ens-meta-prod-1-ensprod',
+    compara_db => '',
+    compara_host => 'mysql-ens-sta-3',
 
     ## Allow division of compara database to be explicitly specified
     compara_division => undef,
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
```diff
@@ -45,6 +45,10 @@ sub default_options {
     run_all      => 0,
     antispecies  => [],
     meta_filters => {},
+    production_db => 'ensembl_production',
+    production_host => 'mysql-ens-meta-prod-1-ensprod',
+    compara_db => '',
```
james-monkeyshines: I think it'd be better to delete this line. If there is no default for compara_db, then if you forget to supply the parameter, it will fail at the init_pipeline step. Otherwise, the first analysis will fail, because the command will not work if an empty string is used.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
```diff
@@ -45,6 +45,10 @@ sub default_options {
     run_all      => 0,
     antispecies  => [],
     meta_filters => {},
+    production_db => 'ensembl_production',
+    production_host => 'mysql-ens-meta-prod-1-ensprod',
```
james-monkeyshines: I don't think this needs write access? Probably best to minimise how much info we provide about our servers and drop the 'ensprod'? In fact, could go a step further and just have `meta1`?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
thomasmaurel: Well spotted, thanks @james-monkeyshines. I've removed the line now.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneTreeHighlighting_conf.pm
thomasmaurel: Well spotted @james-monkeyshines, I got it wrong. production host can be ro but compara_host need to br rw. I've updated the code and use short names too.

PR #296, opened 2019-07-02
Blat update

thomasmaurel:

## Requirements

## Description

The Web team only have BLAT servers for 16 species and human GRCh37. Generating these files for all the species is not really useful since the files are not used by the Web team and external people cannot really use them too. This change only generate Blat files for the 16 species and add the VM port number at the start of the file name. More details here: https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-597

## Use case

Generate BLAT files for the Web team

## Benefits

We will only update BLAT files if assembly has changed or is updated for a small subset of species

## Possible Drawbacks

None.

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/FASTA/BlatIndexer.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/FASTA/BlatIndexer.pm
@@ -75,26 +75,6 @@ use File::stat;
 use Bio::EnsEMBL::Utils::IO qw/work_with_file/;
 use Bio::EnsEMBL::Utils::Exception qw/throw/;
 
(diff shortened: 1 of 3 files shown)


thomasmaurel:
> Can you confirm that the files are not dumped anymore in the FTP synced directory? I can only see ('ftp_dir') params. If this dir is set to release-XX dump dir... so the files with port name added will be on ftp as well, right? (I presume I am wrong)

Yes @marcoooo I can confirm that. You are right, this is why I am stripping out the release and division here: https://github.com/Ensembl/ensembl-production/pull/296/files#diff-0680a52a81b9c351ae1ad8803f785915R180
Not super elegant but does the work.

PR #328, opened 2019-09-10
Increased the max allowed packed since it failed for a vert species. …

thomasmaurel:
…Removed -n 4 options as it's not really needed here and will make getting onto the queue longer. Removed very high memory as we won't need it

## Requirements

## Description

monodelphis_domestica_core_98_1 needed more packets, so increased the value from 500MB to 1GB.
Cleaned up the memory queues, removed very high memory and -n 4 since we don't need it and make harder to get onto queues.

## Use case

To be able to dump monodelphis_domestica_core_98_1 for 98

## Benefits

We can now dump monodelphis_domestica_core_98_1

## Possible Drawbacks

None.

## Testing

Dependencies
------------


thomasmaurel:
> You changed the resource class but did not update the related pipeline steps. Why?

Because that's not the issue at the end. We have 4GB of memory at the moment and the job that used the more memory was 1GB. We need to investigate this further.

PR #344, opened 2019-10-03
Fixed issues where checksum_generator analysis was trying to generate…

thomasmaurel:
… CHECKSUMS for BLAST even when skip_blast option was on. Also fixed issues were convert_fasta for non-vert was running before FASTA dna and PEP were done. I introduced a new flag called skip_convert_fasta so that we don't generate the BLAST FASTA for non-vert when running the pre-dumps or when we don't want to run them. Updated the way skip_blat and skip_ncbiblast were working to only run the BLAST analysis if these flags are off. At the moment the flag is passed to these modules and then no file is generated which looks confusing. Re-worked the flow to make all this to work. Fixed bug introduced with previous commits regarding BLAT, which was creating an empty dir in the release directory. I overwrote the index_path sub for BLAT to make sure BLAT files are generated outside the release directory. Updated oreochromis_niloticus BLAT port number from 30072 to 30073 since there is an assembly update for this species and Web will need to setup a new machine for it.

## Requirements

## Description

Fixed issues where checksum_generator analysis was trying to generate CHECKSUMS for BLAST even when skip_blast option was on. Also fixed issues were convert_fasta for non-vert was running before FASTA dna and PEP were done. I introduced a new flag called skip_convert_fasta so that we don't generate the BLAST FASTA for non-vert when running the pre-dumps or when we don't want to run them. Updated the way skip_blat and skip_ncbiblast were working to only run the BLAST analysis if these flags are off. At the moment the flag is passed to these modules and then no file is generated which looks confusing. Re-worked the flow to make all this to work. Fixed bug introduced with previous commits regarding BLAT, which was creating an empty dir in the release directory. I overwrote the index_path sub for BLAT to make sure BLAT files are generated outside the release directory. Updated oreochromis_niloticus BLAT port number from 30072 to 30073 since there is an assembly update for this species and Web will need to setup a new machine for it.

## Use case

Mainly when running pipeline without BLAST and BLAT but also fixed BLAT issues.

## Benefits

Pipeline will be more configurable. 

## Possible Drawbacks

None
## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/ChksumGenerator.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/ChksumGenerator.pm
@@ -107,13 +107,15 @@ sub get_dirs {
   }
   #Taking care of blast_fasta for non-vertebrates
   if ($self->division() ne "vertebrates"){
-    my $mc = $self->get_DBAdaptor()->get_MetaContainer();
-    if ( $mc->is_multispecies() == 1 ) {
-      my $collection_db = $1 if ( $mc->dbc->dbname() =~ /(.+)\_core/ );
-      push @dirs, File::Spec->catdir($base_path, 'blast_fasta', $self->division(), $collection_db, $species);
-    }
-    else{
-      push @dirs, File::Spec->catdir($base_path, 'blast_fasta', $self->division(), $species);
+    if (not $self->param('skip_convert_fasta')){
+      my $mc = $self->get_DBAdaptor()->get_MetaContainer();
+      if ( $mc->is_multispecies() == 1 ) {
+        my $collection_db = $1 if ( $mc->dbc->dbname() =~ /(.+)\_core/ );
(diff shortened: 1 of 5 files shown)


thomasmaurel:
> I think this is all fine, but the logic here is now very complicated - I couldn't work out what was going on without initialising pipelines with different parameters. I think this is mostly a consequence of the historical accretion of different bits of functionality. But there is now logic for controlling the pipeline in lots of different places: within modules (e.g. ChksumGenerator), in the DumpCore_conf `pipeline_analyses` function, in `tweak_analyses` functions in inheriting conf files, and in conditional flow statements (and sometimes the latter are conditionally added via `tweak_analyses`...)
> I think we will soon need to create a pipeconfig (or several complementary ones) from scratch with a clear set of use cases in mind.

@james-monkeyshines: I agree. I think the complexity is coming from the different BLAST procedures that we have between Vert and non-vert. If Web can settle on one (generated by us or by EBI) then I think the pipeline would simplify a lot. I think you are right, we could have a config file for pre-dumps and for normal run.

PR #374, opened 2020-01-17
DumpCore_conf.pm: use a local copy of xref_LOD_mapping.json

mkszuba:
## Description

Make DumpCore_conf.pm use a local copy of xref_LOD_mapping.json

## Use case

The file in question used to come from VersioningService, which repository is now being phased out.

## Benefits

Do not depend on a deprecated repository. Indeed, reduce the number of external dependencies by one.

## Possible Drawbacks

* A new top-level directory _conf_ has been created to hold the file in question.
* The schema of xref_LOD_mapping.json conforms to JSON Schema draft-07 rather than the current draft/2019-09 (draft-08 in the old nomenclature) due to the fact the validation tools I know do not support the latter yet.

## Testing

Dependencies
------------

Should make the whole ensembl-production no longer depend on VersioningService.

### DIFF
--- a/conf/xref_LOD_mapping.json
+++ b/conf/xref_LOD_mapping.json
@@ -1,0 +1,1690 @@
+{
+  "mappings": [
+    {
+      "db_name": "ArrayExpress",
+      "example_id": "ENSG00000265094",
+      "ignore": true,
+      "bidirectional": false,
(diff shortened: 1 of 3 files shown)


thomasmaurel:
Travis failures are because of copyrights, nothing to do with this PR

PR #382, opened 2020-02-06
Incomplete data file funcgen ensprod 4849

vinay-ebi:
Motif and dna-methy datafiles are missing in ftp release directory. Fixed by adding two adaptors DNAMethylationFileAdaptor and MotifFeatureFileAdaptor.


vinay-ebi:
Only DNA- meth  code implemented. Motif code is commented as API for motif not working

PR #387, opened 2020-02-28
Production db sync ensprod 4852

vinay-ebi:
## Description

When synchronising from the production db to core/core-like/funcgen/variation dbs, it's highly recommended to run datachecks afterwards, to check that the sync has a) worked, and b) not introduced any errors. Production Sync DB hive pipeline  will manage to run the controlled tables, analysis descriptions and applies the relevant datachecks.


vinay-ebi:
> @james-monkeyshines @vinay-ebi Is it possible to think of adding test for this?

james-monkeyshines:
It's _possible_ to do tests, but I'm not sure it's worth the effort in this case. Since the pipeline includes datachecks to verify that the sync has happened, tests on the sync module wouldn't give you much more information. Our (unwritten...) SOP for updating pipeline modules is to subsequently create a test instance of the pipeline to run, so problems should be flagged up there.
The other element to test is the pipeline structure; again, that's do-able, but typically we don't do that, because it's quite time-consuming and complicated, because we don't do it much - obviously it would get easier if we did it more... Here, the pipeline moderately complex, but it doesn't feel like there's enough there to warrant pipeline tests.

james-monkeyshines:
I tweaked too much of the code to be a fair reviewer, so I'm excusing myself from approval...

PR #394, opened 2020-03-16
Fix for RDF dumping losing connection

james-monkeyshines:
## Description
Add disconnectt_when_idle command for RDF dumping, which was losing the DB connection because it spends so long generating vast amounts of data...

## Use case
RDF dump jobs were failing after running for hours, due to a lost DB connection.

## Benefits
Jobs don't fail!

## Possible Drawbacks
None I'm aware of.

## Testing
Tested on dumps for release 100 - half a dozen failures without the fix, none after it. (Could be coincidence, of course, but hard to fully test without significant effort.)


thomasmaurel:
oops sorry @james-monkeyshines. I clicked too quickly on the merge button.

james-monkeyshines:
No worries @thomasmaurel , clicking that big green button can be instinctive...

PR #407, opened 2020-04-21
updated interproscan version

vinay-ebi:
Update InterProScan Version to 5.42.78


thomasmaurel:
@vinay-ebi: I think panther version needs to be updated here: https://github.com/Ensembl/ensembl-production/blob/release/101/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProteinFeatures_conf.pm#L109 as I think you said that you updated it to 15.

PR #410, opened 2020-04-23
Update gene and transcript query to exclude genes/transcripts that ar…

thomasmaurel:
…e not displayed on the ensembl website. e.g we have some human protein in the sheep otherfeatures database (see here for details: ENSPROD-5362). Also fixed the gene auto complete code.

## Requirements

## Description

Update gene and transcript query to exclude genes/transcripts that are not displayed on the ensembl website. e.g we have some human protein in the sheep otherfeatures database (see here for details: ENSPROD-5362). Also fixed the gene auto complete code.

## Use case

This will mainly affects the Gene search dumps for the otherfeatures database.

## Benefits

We won't have Human proteins in the sheep EBI search dumps anymore

## Possible Drawbacks

None

## Testing

Dependencies
------------


thomasmaurel:
> The conditions are perhaps a little loose in theory, i.e. allowing any characters between "gene" and "do_not_display", means that they are not necessarily associated with each other. But in practice this will always work with our current set of tracks, and I don't think additional complexity is warranted.

@james-monkeyshines: That's a very good point. The tests were failing because I actually missed the real cause of the issue. The problem was that we moved the web_data from Perl to Json but we missed this bit of code here: https://github.com/Ensembl/ensembl-production/pull/410/commits/ba0ecbb3e697bb6e8974539f45c4a71d2a7f6851#diff-50e9157acf2e5f21492fe127858f7d89L81.
I believe this captures the web_data in a better way. I've updated the code accordingly and updated the tests web_data.

PR #411, opened 2020-04-23
Add xref datacheck to gpad

vinay-ebi:
Adding Xref related Datachecks to GPAD pipeline.


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GPAD_conf.pm
```diff
@@ -241,9 +241,41 @@ sub pipeline_analyses {
...
+                              history_file    => $self->o('history_file'),
+                              failures_fatal  => 1,
+                            },
+      -flow_into         => 'RunXrefAdvisoryDatacheck', 
+      -max_retry_count   => 1,
+      -analysis_capacity => 10,
+      -batch_size        => 10,
+    },
+    {
+      -logic_name        => 'RunXrefAdvisoryDatacheck',
```
james-monkeyshines, line 264: Advisory checks often compare data to the previous release's database, so a parameter is required to specify the appropriate server. One way to do this is with `-old_server_uri`, and then have the Confluence SOP pass in the mirror server that is appropriate for each staging server.
But a cleaner way to do it is to use the shared config files in the datacheck/config directory. There's one per staging server, and they already contain the necessary old server details. They also contain standard locations for history and output too, so it's three birds with one stone. For maximum flexibility in calling the pipeline, probably best for config_file to have a default of undef, as with history_file.

james-monkeyshines:
As an aside, I don't know why travis is failing. This branch passes all tests when I run locally...

thomasmaurel:
> As an aside, I don't know why travis is failing. This branch passes all tests when I run locally...

@james-monkeyshines: The push Travis is expected to fail because it's using the repos on master and the core team have already updated to 102 so there is a mismatch.
The PR is supposed to pass, I have restarted it to see, sometimes Travis is a bit flaky.

PR #415, opened 2020-05-05
Add SampleData_vertebrates_conf

radome:

## Requirements

## Description

Add `SampleData_vertebrates_conf` for specific configuration of Sample Data Pipeline for Vertebrates. In particular this adds the `antispecies` parameter for excluding species from the pipeline (default value: homo_sapiens). 

## Use case

Sample Data Pipeline's step `GenerateSampleData` is failing on homo_sapiens, but it shouldn't be ran on that database in the first place as the sample data is manually curated.

## Benefits

Being able to exclude species with the `antispecies` parameter and add further

## Possible Drawbacks

Change SOPs to use `SampleData_vertebrates_conf` instead of `SampleData_conf`

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
@@ -40,19 +40,6 @@ package Bio::EnsEMBL::Production::Pipeline::PipeConfig::SampleData_vertebrates_c
 use strict;
 use warnings;
 use base ('Bio::EnsEMBL::Production::Pipeline::PipeConfig::SampleData_conf');
-use Bio::EnsEMBL::ApiVersion qw/software_version/;
-use Bio::EnsEMBL::Hive::Version 2.5;
-use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf;
-
-
-=head2 default_options
-
- Description: It returns a hashref containing the default options for HiveGeneric_conf
- Returntype : Hashref
- Exceptions : None
-
-
-=cut
 
 sub default_options {
     my ($self) = @_;
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
@@ -1,0 +1,68 @@
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
```diff
@@ -0,0 +1,68 @@
...
+
+
+=head2 default_options
+
+ Description: It returns a hashref containing the default options for HiveGeneric_conf
+ Returntype : Hashref
+ Exceptions : None
+
+
+=cut
```
james-monkeyshines, line 55: I realise that this has just been copied from the nonvert config, but none of this is required - it doesn't hurt, but there's no point in pulling in libraries that you don't need, or having documentation that doesn't really tell you anything useful...

radome: Agree! Fixed in 5784cf8

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
```diff
@@ -0,0 +1,55 @@
...
+use base ('Bio::EnsEMBL::Production::Pipeline::PipeConfig::SampleData_conf');
+
+sub default_options {
+    my ($self) = @_;
+
+    return {
+        # inherit other stuff from the base class
+        %{ $self->SUPER::default_options() },
+	    # species with curated sample genes
+        antispecies   => ['homo_sapiens'],
```
thomasmaurel, line 51: We might need more species here like mouse, rat... @james-monkeyshines do you know the full list or should we check with the genebuild team?

thomasmaurel:
@radome: Once this gets merged, don't forget to update the conf file in the documentation here: https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Generate+Sample+Data#GenerateSampleData-Vertebrates

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_vertebrates_conf.pm
james-monkeyshines: Yes, agree about mouse and rat, probably zebrafish too... Need to get genebuild to give us a full ist too. Or maybe it would be better to do this only for new species, not updated ones. (If an geneset update removed the sample gene a datacheck would fail, so that would be flagged up before handover to us.)

thomasmaurel: @leannehaggerty: Could you please give us a list of Vertebrates species for which we don't want to overwrite the sample data?

leannehaggerty: Sorry I missed this. These are my thoughts, but happy to provide a list if you think necessary:
I think James' suggestion to run for all new species is probably the best solution - that way everything that already exists will stay intact. 
In the event of a gene set update we (Genebuild) will need to check before handover that the sample gene still exists in the gene set.
And finally if there is a specific sample gene that needs to be set for a species in the future, i.e. we don't want to use the sample gene script to update it, then we should add this species to the 'antispecies' list.

leannehaggerty: Hi all

Have discussed with the Genebuilders and think the initial list should be human, mouse, zebrafish, pig, sheep, horse and cow. 
However, for the sheep that has just been handed over for e101 (sheep (rambouillet)), the script will need to be run, just this once. 
Does this sound ok? 

When new species need to be added to the list would you prefer us to update SampleData_vertebrates_conf.pm via a PR or is there a static list we should add to?

Cheers

PR #418, opened 2020-05-11
Add xref datacheck to pipeline

vinay-ebi:
Adding Xref related Datachecks to Gene Description projection and xref pipeline.


vinay-ebi:
PR 419  covers all the xref datachecks  commits.

PR #425, opened 2020-05-29
Adding Update Compare gene name and description to the gene name/desc…

thomasmaurel:
… projection pipeline so that we can run these tasks in one go. The Compara pipeline is run after Email Report. The documentation will need to be updated once we are done running this pipeline for 101. We still need to keep the Compare gene member documentation for Pan since we don't run projections for Pan

## Requirements

## Description

Adding Update Compare gene name and description to the gene name/desc projection pipeline so that we can run these tasks in one go. The Compara pipeline is run after Email Report. The documentation will need to be updated once we are done running this pipeline for 101. We still need to keep the Compare gene member documentation for Pan since we don't run projections for Pan

## Use case

We run this pipeline every release

## Benefits

We need to run the compara pipeline after projections so this save us some time

## Possible Drawbacks

The Compara gene name updates hive config won't be updated/maintained by Compara anymore. They have not changed it in the last two years so should be ok. They still maintain the modules.

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneNameDescProjection_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneNameDescProjection_conf.pm
@@ -90,6 +90,7 @@ sub default_options {
     history_file   => undef,
     config_file    => undef,
     old_server_uri => undef
+
   };
 }
 
@@ -115,6 +116,7 @@ sub pipeline_wide_parameters {
   return {
     %{$self->SUPER::pipeline_wide_parameters},
     'store_projections' => $self->o('store_projections'),
+    'compara_db' => $self->o('compara_db')
   };
 }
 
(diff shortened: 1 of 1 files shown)


thomasmaurel:
Screenshot 2020-05-29 at 16 54 15

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/GeneNameDescProjection_conf.pm
```diff
@@ -374,6 +376,60 @@ sub pipeline_analyses {
                             subject => $self->o('pipeline_name').' has completed',
                             text    => 'Log files: '.$self->o('output_dir'),
                           },
+      -flow_into => [ 'species_update_factory' ],
+    },
+    {
```
james-monkeyshines, line 381: This is almost exactly what is in Bio::EnsEMBL::Compara::PipeConfig::Parts::UpdateMemberNamesDescriptions - so rather than replicate (and maintain) it, could instead have: `@{Bio::EnsEMBL::Compara::PipeConfig::Parts::UpdateMemberNamesDescriptions::pipeline_analyses_member_names_descriptions($self)}`
with a default parameter value for `update_capacity`, and a resource_classes function that inherits from the parent and adds `1Gb_job`. (`compara_db` is set as a pipeline wide parameter, so doesn't need to be passed explicitly.)

PR #430, opened 2020-06-05
[ENSCOMPARASW-3527] Allow division-prefixed Ancestral sequences Registry entries

muffato:
## Use case

See ENSWEB-5706 and related tickets for more information. The REST server sees all MySQL servers (Vertebrates and non-Vertebrates), which contain in total two ancestral databases, but the current Registry model only allows one. The Compara API then fails because it is not able to access the other one

## Description

I am proposing to register the ancestral databases with their division name, preventing a clash between the Plants and Vertebrates databases. The change is being made in the Core repository, and ensembl-production has to be updated because several modules compare the registry name to "Ancestral sequences", which will now be "Plants Ancestral sequences" for Plants.

Two questions:
1. Do you still maintain this branch (`release/100`) ? I can make the PR onto `release/101` instead
2. Do you have a mechanism to merge this onto the more recent branches ?

Anyway, the later branches should be reviewed because there could be new modules that check the registry name and need to be updated.

:warning: _Do not merge until *all* these related pull-requests have been approved !_: Ensembl/ensembl#488, Ensembl/ensembl-compara#197

Equivalent pull-requests: #432 for `release/101` and #433 for `master`

## Benefits

Production code will still work on the NV server, and will also work of all databases are mixed together (physically or via a Registry configuration file).

## Possible Drawbacks

None

## Testing

No

All good

Yes, once the branch for all the `git clone` is forced to be `release/100`, see https://travis-ci.org/github/Ensembl/ensembl-production/builds/695944537

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbFactory.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Common/DbFactory.pm
@@ -123,7 +123,7 @@ sub run {
 
   if ($run_all) {
     foreach my $dba (@$all_dbas) {
-      unless ($dba->species eq 'Ancestral sequences') {
+      unless ($dba->species =~ /Ancestral sequences/) {
         $self->add_species($dba, \%dbs);
       }
     }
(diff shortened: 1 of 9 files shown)


james-monkeyshines:
I see no harm in merging this now, rather than waiting for the ensembl and ensembl-compara PRs - the code will continue to work for the extant situation, abut would also be ready for the proposed changes. And it makes life a bit easier to have it merged before we branch for 102, which is imminent.

muffato:
Yes you're right, I was probably too cautious :) I kept it compatible with the "old style naming" just in case.
Thanks !

PR #431, opened 2020-06-09
dbname parameter added to syncDB pipeline to sync controll tables and…

vinay-ebi:
SyncDB pipeline to sync a named database, rather than using species/division parameters.


vinay-ebi:
> @vinay-ebi Is this enough to actaully enable "dbname" only updates ? I would have expect some updates in modules PopulateControlledTables and PopulateAnalysisDescription.




There is already an option in DbFactory module, we just need to add it to the conf file

PR #444, opened 2020-06-29
ENSCORESW-3196: separate checksum entries by source to avoid cross so…

magaliruffier:
…urce mappings

## Requirements

## Description

We have checksum entries for RNAcentral (transcript sequences) and UniParc (protein sequences)
Currently, we do not distinguish between the two sources and attempt to map checksums from both sources to all sequences available.
Recently, due to clash in checksums, we have started mapping some UniParc checksums to transcript sequences.
To avoid this, we want to store the checksums with the source they come from and use that information in the later mapping stage, to only attempt to map RNAcentral to transcript sequences, and UniParc to protein sequences

## Use case

UPI0001765128 is a UniParc entry with checksum CFECF93F30021B262B430492E53B847C
The transcript sequence for ENST00000616594.2 has the same checksum and maps to UPI0001765128
With this change, UPI0001765128 is only compared to protein sequences and will not map to ENST00000616594.2

## Benefits

No cross-source checksum mapping happening

## Possible Drawbacks

The individual checksum mapping stages might be slower, as the SQL query restricts on individual source_id

## Testing

NA
NA
NA
There is no test case for this pipeline
The whole xref pipeline was run before and after the change.
In particular, the DataChecks run at the end were reporting the mismapped UPI0001765128 xref before the change, but not after
NA

Dependencies
------------

To fully work, this will also require the following change in the Ensembl repository: https://github.com/Ensembl/ensembl/pull/496

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
@@ -85,6 +85,9 @@ sub download_file {
       if ( !match_glob( basename( $uri->path() ), $remote_file ) ) { next; }
       $remote_file =~ s/\///g;
       $file_path = catfile($dest_dir, basename($remote_file));
+      if (defined $db and $db eq 'checksum') {
+        $file_path = catfile($dest_dir, $source_name."-".basename($remote_file));
+      }
       $ftp->get( $remote_file, $file_path ) unless $skip_download_if_file_present and -f $file_path;
     }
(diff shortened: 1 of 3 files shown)


thomasmaurel:
@magaliruffier: Does this need to go into release 102 because we are going to run the xrefs very soon.

magaliruffier:
it is not an urgent fix, fine to keep for release 103
release 102 has already been branched, I would not dare suggest a post-branching fix ;-)

PR #446, opened 2020-07-04
Features/rr convert meta 2 hive species

marcoooo:

## Description

For Manual RR - A script to generated a list of species parameters from metadata report_genomes.pl outputs files

## Use case

RR retrieve expected species to proceed in RR.

## Benefits

No manual edition of species list per RR

## Possible Drawbacks

None

## Testing

Dependencies
------------

### DIFF
--- a/scripts/py/meta_hive_species.py
+++ b/scripts/py/meta_hive_species.py
@@ -1,16 +1,19 @@
 #!/usr/bin/env python
 """
-.. See the NOTICE file distributed with this work for additional information
-   regarding copyright ownership.
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-       http://www.apache.org/licenses/LICENSE-2.0
-   Unless required by applicable law or agreed to in writing, software
-   distributed under the License is distributed on an "AS IS" BASIS,
-   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-   See the License for the specific language governing permissions and
-   limitations under the License.
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
 """
 
 '''
--- a/scripts/py/meta_hive_species.py
+++ b/scripts/py/meta_hive_species.py
@@ -66,7 +66,6 @@ if __name__ == '__main__':
     if not args.output_file:
         logger.setLevel(logging.ERROR)
     if args.division:
-        assert args.division in divisions, "Unknown Division name, please chose in {}".format(divisions)
         logger.info("Process only division %s", args.division)
         divisions = [args.division]
     else:
--- a/scripts/py/meta_hive_species.py
+++ b/scripts/py/meta_hive_species.py
@@ -39,8 +39,8 @@ parser = argparse.ArgumentParser(
 parser.add_argument('-i', '--input_dir', type=str, required=True, help='Input directory')
 parser.add_argument('-d', '--division', help='Restrict to a single division', type=str, choices=divisions)
 parser.add_argument('-f', '--output_file', help='Overwrite default output file name', required=False)
-parser.add_argument('-t', '--genomes_types', help='List genomes events (n,r,ua,ug)', required=False, type=str,
-                    default='n,r,ua,ug')
+parser.add_argument('-t', '--genomes_types', help='List genomes events (n r ua ug)', type=str, nargs='+',
+                    default=['n', 'r', 'ua', 'ug'])
 parser.add_argument('-o', '--output', help='species,dbname', required=False, type=str,
                     default='species')
 
--- a/scripts/py/meta_hive_species.py
+++ b/scripts/py/meta_hive_species.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+"""
+.. See the NOTICE file distributed with this work for additional information
+   regarding copyright ownership.
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+"""
+
+'''
+This script takes an directory path as parameter and convert the output from ~metadata/misc_scripts/report_genomes.pl
+and turn it into a flat file names `species_list.txt` to `cat` into a eHive pipeline configuration script.
+Example:
+- species AAA \
+- species BBB \
+     
+'''
+import argparse
+import csv
+import logging
+from os import getenv
+from os.path import isdir
+from os.path import join, isfile, realpath
+
+logging.basicConfig(level=logging.INFO, format='%(message)s')
+
+parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+
+parser = argparse.ArgumentParser(
+    description='Parse outputs from `~ensemble-metadata/misc_scripts/genomes_report.pl to get species_list for hive'
+                ' init_pipeline command')
+parser.add_argument('-i', '--input_dir', type=str, required=True, help='Input directory')
+parser.add_argument('-d', '--division', help='Restrict to a single division', type=str)
+parser.add_argument('-f', '--output_name', help='Overwrite default output file name', required=False, type=str,
+                    default='species.txt')
+parser.add_argument('-t', '--genomes_types', help='List genomes events (n,r,ua,ug)', required=False, type=str,
+                    default='n,r,ua,ug')
```
radome, line 43: Unless you want the user to specify the genome types as a comma separated values in a single string, you can use this handy `argparse` argument for treating lists:
```suggestion
parser.add_argument('-t', '--genomes_types', help='List genomes events (n r ua ug)', type=str, nargs='+',
                    default=['n', 'r', 'ua', 'ug'])
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+    home_dir = getenv("HOME")
+    assert isdir(args.input_dir), "Input dir is mandatory when not running report_genomes.pl"
+    scan_dir = realpath(args.input_dir)
+    logging.info("Input Dir: {}".format(realpath(scan_dir)))
+
+    species = []
+    expected_files_names = [names[n] for n in args.genomes_types.split(',')]
+
+    for div in divisions:
+        logging.info("Division: {}".format(div))
```
radome: Lazy logging:
```suggestion
        logging.info("Division: %s", div)
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+        expected_files = [ ex.format(div) for ex in expected_files_names]
+        for file in expected_files:
+            file_path = join(scan_dir, file)
+            if isfile(file_path):
+                with open(file_path, 'r') as f:
+                    logging.info("Current File: {}".format(f.name))
```
radome: Lazy logging:
```suggestion
                    logging.info("Current File: %s", f.name)
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+                    reader = csv.reader(f, delimiter="\t")
+                    next(reader)
+                    for row in reader:
+                        species.append(row[0])
+            else:
+                logging.info("File not found: {}".format(file_path))
```
radome: Lazy logging:
```suggestion
                logging.info("File not found: %s", file_path)
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+
+    logging.info("Retrieved species:")
+    logging.info(species)
+    out_file_path = join(home_dir, args.output_name)
+    with open(out_file_path, 'w') as f:
+        for spec in species:
+            f.write('-species {} '.format(spec))
+    logging.info("File generated in {}".format(out_file_path))
```
radome: Lazy logging:
```suggestion
    logging.info("File generated in %s", out_file_path)
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,107 @@
...
+        logger.setLevel(logging.ERROR)
+    if args.division:
+        assert args.division in divisions, "Unknown Division name, please chose in {}".format(divisions)
+        logger.info("Process only division {}".format(args.division))
+        divisions = [args.division]
+    else:
+        logger.info("Process all divisions")
+
+    home_dir = getenv("HOME")
+    assert isdir(args.input_dir), "Input dir is mandatory when not running report_genomes.pl"
```
radome, line 73: I'd suggest exit the program nicely here instead of crashing out with an `assert`:
```suggestion
    if not isdir(args.input_dir):
        sys.exit("Input dir is mandatory when not running report_genomes.pl")
```
If you implement this, remember to `import sys`.
More on `sys.exit` https://docs.python.org/3/library/sys.html#sys.exit

In general I'd suggest against the use of `assert` if not for checking invariants.
More on assertions: https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/

radome:
Also I'd suggest using lazy logging interpolation in general, as it evaluates the parameters only if the log it's actually executed (it could be prevented by the log level).

Example:
```python
# Lazy logging interpolation
logger.info("This is my object: %s", object)

# Non lazy logging interpolation
logger.info("This is my object: {}".format(object))
```

scripts/py/meta_hive_species.py
```diff
@@ -0,0 +1,98 @@
...
+   'n': '{}-new_genomes.txt',
+   'r': '{}-renamed_genomes.txt',
+   'ua': '{}-updated_annotations.txt',
+   'ug': '{}-updated_assemblies.txt',
+}
+
+if __name__ == '__main__':
+    args = parser.parse_args()
+    if args.division:
+        assert args.division in divisions, "Unknown Division name, please chose in {}".format(divisions)
```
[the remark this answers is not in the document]
marcoooo: Didn't know about the "List" feature, thanks !

marcoooo:
> Also I'd suggest using lazy logging interpolation in general, as it evaluates the parameters only if the log it's actually executed (it could be prevented by the log level).
> 
> Example:
> 
> ```python
> # Lazy logging interpolation
> logger.info("This is my object: %s", my_object)
> 
> # Non lazy logging interpolation
> logger.info("This is my object: {}".format(my_object))
> ```

I would have thought the '%' notation was the "oldschool" version, and '.format' was the new way to format string.

marcoooo:
> > Also I'd suggest using lazy logging interpolation in general, as it evaluates the parameters only if the log it's actually executed (it could be prevented by the log level).
> > Example:
> > ```python
> > # Lazy logging interpolation
> > logger.info("This is my object: %s", my_object)
> > 
> > # Non lazy logging interpolation
> > logger.info("This is my object: {}".format(my_object))
> > ```
> 
> I would have thought the '%' notation was the "oldschool" version, and '.format' was the new way to format string.

From https://docs.python.org/3.4/library/stdtypes.html#printf-style-string-formatting

Note

The formatting operations described here exhibit a variety of quirks that lead to a number of common errors (such as failing to display tuples and dictionaries correctly). Using the newer str.format() interface helps avoid these errors, and also provides a generally more powerful, flexible and extensible approach to formatting text.

radome:
> > > Also I'd suggest using lazy logging interpolation in general, as it evaluates the parameters only if the log it's actually executed (it could be prevented by the log level).
> > > Example:
> > > ```python
> > > # Lazy logging interpolation
> > > logger.info("This is my object: %s", my_object)
> > > 
> > > # Non lazy logging interpolation
> > > logger.info("This is my object: {}".format(my_object))
> > > ```
> > 
> > 
> > I would have thought the '%' notation was the "oldschool" version, and '.format' was the new way to format string.
> 
> From https://docs.python.org/3.4/library/stdtypes.html#printf-style-string-formatting
> 
> Note
> 
> The formatting operations described here exhibit a variety of quirks that lead to a number of common errors (such as failing to display tuples and dictionaries correctly). Using the newer str.format() interface helps avoid these errors, and also provides a generally more powerful, flexible and extensible approach to formatting text.

I'm not suggesting formatting log strings using the `%-format` operator like this:
```python
logger.info("This is my object: %s" % my_object)
```
I'm suggesting following this https://docs.python.org/3.5/howto/logging.html#logging-variable-data which has the advantage of transforming the arguments passed to the logger into strings only when needed.

So like this:
```python
logger.info("This is my object: %s", my_object)
```

In our particular cases here isn't a massive deal of computation actually, it's more to be consistent with the intended usage of `logging.debug()` https://docs.python.org/3.7/library/logging.html#logging.debug. Also we won't incur in the quirks of `%-format` operator (used by the `logging` module internally) as we're simply using the `%s` format which is safe in this context.

For other cases, I agree, and prefer `.format()` as for the general case is safer and more powerful. For example:
```python
# Tuples will be interpreted as multiple arguments
tup = ('a', 1, 'b')
'%s' % tup  # Raises TypeError: not all arguments converted during string formatting
'{}'.format(tup)  # Returns "('a', 1, 'b')"

# Lists access doesn't work as expected
lst = ['a', 'b', 'c']
'%(0)s %(2)s' % lst # Raises TypeError: list indices must be integers or slices, not str
'{0} {2}'.format(*lst)  # Returns "a c". Note that the %-format operator does not support unpacking
```

scripts/py/meta_hive_species.py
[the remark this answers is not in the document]
radome: This `assert` is redundant now as `argparse` will already prompt the users if they specify a `division` not in `divisions` as we set `choices` for that argument.

PR #447, opened 2020-07-08
Add dcstat script for parsing DC JSON results

radome:

## Requirements

Python 3.7+

## Description

Console Python script for parsing and reporting data from a Datacheck JSON report.

## Use case

In many cases DCs generate lots of failures, some of them might not be relevant. This scripts help filtering out the not relevant ones and potentially add more complex criteria.

### DIFF
--- a/scripts/py/dcstat.py
+++ b/scripts/py/dcstat.py
@@ -1,5 +1,18 @@
 #!/usr/bin/env python
 
+# Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+# Copyright [2016-2020] EMBL-European Bioinformatics Institute
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
 # Requirements:
 #   Python 3.7+
 # Usage example:
@@ -10,9 +23,16 @@
 
 import argparse
 from functools import partial
+import logging
 import json
 import re
 from typing import NamedTuple, List, Pattern, Callable
+
+
+logging.basicConfig(format='%(levelname)s: %(message)s')
+
+
+logger = logging.getLogger(__name__)
 
 
 class FailedTests(NamedTuple):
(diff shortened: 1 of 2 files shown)


scripts/py/dcstat.py
```diff
@@ -152,17 +207,22 @@ def main():
             help='Verbosity level: 1) shows only DB names. 2) shows also failed tests names. 3) shows also errors log')
     parser.add_argument('-s', '--filter-species', type=str, nargs='+',
                         help='Filter results by species.')
+    parser.add_argument('-Q', '--quiet', action='store_true',
```
marcoooo, line 210: good reason or typo in capitalized Q?

scripts/py/dcstat.py
```diff
@@ -1,5 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+
 # Requirements:
 #   Python 3.7+
```
marcoooo, line 18: Default version on farm is 3.5.1, is it not too early to prepare a script for 3.7? Or has it been approved to move to 3.7 as a default already? Is it sensible to stop the script if python version is too low?

scripts/py/dcstat.py
radome: No, that's intended. I use upper case for options that do not require an argument, and lower case for the argument ones. I'm not sure where that habit comes from actually. Maybe common GCC flags? I don't know.

marcoooo: Well ok, first time I see this "rule" :-D no problem.

scripts/py/dcstat.py
radome: Python 3.7.6 is installed and working on the farm already. I think new projects should start from that version unless there are good reasons not to (like libraries that absolutely need to be 3.5 compatible) otherwise we will never move forward and 3.5 is reaching is end of life anyway.

Moreover this is a standalone command line script and doesn't have dependencies (i.e no need for a virtualenv to run it) so it can be run on the farm by simply:
```
$ pyenv activate 3.7.6
$ ./dcstat.py ...
```

marcoooo: Well Was not expecting this script to need a virtual env, hence my question.

marcoooo: Since when running pyenv, in the background you are activating a venv, right?

radome: Yes, you're activating the default one for that version. What I meant is that there is no need for creating a specific virtualenv for this script as it will work with any Python 3.7+ so it can work immediately in the farm without any setup action from admins, but switching to Python 3.7.6 is required to the user.

marcoooo:
Approved, We have to think about raising Python default version on farm to 3.6 at some point.

PR #448, opened 2020-07-09
Use datacheck groups relevant to the pipeline

james-monkeyshines:
## Description
Multiple xref-related pipelines were using the same 'xref' datacheck group - some of these are specific to the work of a particular pipeline, so running them in another pipeline is redundant and potentially confusing.

## Benefits
Pipelines only run relevant datachecks.

## Possible Drawbacks
None I can think of.

## Testing
None done, but this is a very simple update.


james-monkeyshines:
This pairs with https://github.com/Ensembl/ensembl-datacheck/pull/272

PR #450, opened 2020-07-15
Remove xref from external db

vinay-ebi:
Removed xref record from external db for  non human


vinay-ebi:
PR need raised against master. closing the PR

PR #467, opened 2020-09-02
Deprecate RDF dumps

james-monkeyshines:
## Description
Starting from Ensembl Release 103, we will not be producing dumps in RDF format. This PR removes the config from the FTP dumping pipeline, and all RDF-associated code. Related intention: https://www.ebi.ac.uk/panda/jira/browse/ENSINT-519


marcoooo (approved):
That's some good cleaning!

PR #473, opened 2020-09-03
Add option for Production DB sync pipeline to send an email on completion

james-monkeyshines:
## Description
Feature request from the Variation team: an option for Production DB sync pipeline to send an email on successful completion, with confirmation of datacheck results (ENSPROD-5422). Previously, datacheck output was always stored in the database, have added the option to send it to a file, so that it's easier to get at.

## Benefits
Provide easy access to results of datachecks, reassurance that history is saved in the appropriate file.

## Possible Drawbacks
None.

## Testing
Pipeline was initialised and run to completion with various parameter configurations.


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProductionDBSync_conf.pm
```diff
@@ -38,6 +43,7 @@ sub default_options {
     populate_controlled_tables    => 1,
     populate_analysis_description => 1,
     group => [],
+    email_on_completion => 0,
```
marcoooo, line 46: I would personally turn it on by default.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProductionDBSync_conf.pm
```diff
@@ -257,7 +284,20 @@ sub pipeline_analyses {
                               pipeline_name => $self->o('pipeline_name'),
                             },
       -rc_name           => 'normal',
-   },
+    },
```
marcoooo, line 287: Quick question: is it sent only when failures? (linked to the "enabled by default" :-))

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProductionDBSync_conf.pm
james-monkeyshines: In general, our pipelines don't email you when they complete - it's okay for us, because we tend to do a small number of pipelines on divisions. But other teams might be running this per-species, and lots of confirmation emails might be annoying, so I prefer to keep it as opt-in.

marcoooo (approved):

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProductionDBSync_conf.pm
james-monkeyshines: No, you only get an email if the pipeline completes successfully. If an advisory datacheck fails, there is already code in the pipeline to send email notifications about those separately. If a critical datacheck fails, the job will fail and the pipeline will not complete until the problem is resolved and the job reset and re-run (or ignored and okayed).

PR #475, opened 2020-09-09
EP-5842: Ontology Pipeline DBCopy fix & terms DataChecks

radome:
## Benefits
- Add specific term DataChecks to Ontology Load pipeline
- Set DataChecks to _critical_
- Separate copy stage to fix the "renaming problem" with DB Copy service
- Automate skip of PHI ontology load when not needed

## Possible Drawbacks

No GUI Hive way to forgive DataChecks and proceed with the copy stage. In that case copy jobs have to be seeded via command line using `seed_pipeline.pl`

Dependencies
------------
This changes: https://github.com/Ensembl/ensembl-datacheck/pull/298

## Testing


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm
```diff
@@ -222,8 +221,8 @@ sub pipeline_analyses {
             -max_retry_count => 1,
             -flow_into       => {
                 1 => WHEN(
-                    '#skip_phi#' => [ 'compute_closure' ],
-                    ELSE [ 'phibase_load_factory' ]
+                    '("PHI" ~~ #ontologies# and (#wipe_one# == 1 or #wipe_all# == 1))' => [ 'phibase_load_factory' ],
```
james-monkeyshines, line 224: I've steered clear of smartmatching in Perl - too smart for my liking, and since it's still an experimental feature, the warnings are annoying, if nothing else...
An alternative is to use Perl's `grep` method, which I've always thought was a misleading use of nomenclature... haven't tested it in the context of this pipeline, but the following should do the same thing, I think:
 ```suggestion
                    '( #expr(grep {/PHI/} @{#ontologies#})expr# and (#wipe_one# == 1 or #wipe_all# == 1) )' => [ 'phibase_load_factory' ],
```

File: modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm [hunk 2]
```diff
@@ -33,7 +33,7 @@ package Bio::EnsEMBL::Production::Pipeline::PipeConfig::OLSLoad_conf;
 
 use strict;
 use base ('Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf');
-use warnings FATAL => 'all';
+# use warnings FATAL => 'all';
```
marcoooo, line 36: Is it something intended? or did you deactivate it for testing purpose?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm
```diff
@@ -33,7 +33,7 @@ package Bio::EnsEMBL::Production::Pipeline::PipeConfig::OLSLoad_conf;
 
 use strict;
 use base ('Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf');
```
james-monkeyshines, line 35: You could consider using 'Bio::EnsEMBL::Production::Pipeline::PipeConfig::Base_conf' instead. That inherits from the HiveGeneric_conf, but adds in some useful stuff, to say redefining things. In this case, there are 'user' and 'email' parameters already defined, so you can just use `$self->o('user')` and `$self->o('email')` in the payloads; and also an 'ensembl_release' parameter, which defaults to the version of your local ensembl repo, that could replace 'ens_version'.

james-monkeyshines (commented):
Everything here works fine, but I've made a couple of suggestions.
Not related to the changes here, but I noticed that the resource classes have `-n 4` - I suspect that this is a consequence of copy-and-pasting, rather than a need for all pipeline jobs to use 4 cores? Removing this constraint should mean that the high mem jobs get farm nodes more quickly.

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm
marcoooo: I guessed this was the reason behind the inactivation of fatal?

radome: I've tried with that syntax but it doesn't seems to be working, as I get the following error:
```
ParamError: Cannot evaluate the expression: 'expr(( #expr(grep {/PHI/} @{#ontologies#})expr# and (#wipe_one# == 1 or #wipe_all# == 1) ))expr' ==> '( #expr(grep {/PHI/} @{$self->_param_possibly_overridden('ontologies', $overriding_hash)})expr# and ($self->_param_possibly_overridden('wipe_one', $overriding_hash) == 1 or $self->_param_possibly_overridden('wipe_all', $overriding_hash) == 1) )' syntax error at (eval 426) line 2, at EOF
```

james-monkeyshines:
```suggestion
                    '(grep {/PHI/} @{#ontologies#} and (#wipe_one# == 1 or #wipe_all# == 1))' => [ 'phibase_load_factory' ],
```
Sorry, got the syntax wrong, didn't need the #expr because that's already implied by the enclosing parentheses.

james-monkeyshines (commented):
When I created a sample pipeline, to work out why the PHI-grep-bit wasn't working, it stumbled over the 'init_meta' analysis. Maybe something's not right in my environment, but it seems like the OLSHiveLoader module isn't meant to be used directly, rather as a base class to inherit from?

marcoooo (approved):

PR #477, opened 2020-09-15
Add duplicate report to Stable ID pipeline

james-monkeyshines:
## Description
We are not supposed to have duplicate stable IDs either within or across species. We can make sure of the former with the GeneStableID datacheck, but the latter is not enforcable, because we are not in control of the stable IDs that are used in the gene sets that we import, rather than annotate in-house.

## Use case
We want to have a report of duplicate stable IDs, so that we can a) check for egregious errors, and b) understand the scale and nature of duplication, and even if it can't easily be tackled we can provide this information to those using the Ensembl data, so that they can account for it if necessary.

There was a nascent datacheck for this, but it was only comprehensive if you provided the right parameters; it was easy to get a passing result by accidentally omitting a database server. It also used colossal amounts of memory, and it didn't really fit into the processes we have for running datachecks.

Since we create, every release, a database with all stable IDs in an indexed table, it makes sense to use that to get a comprehensive report on the duplicates. This PR adds such a step to the stable ID pipeline (as well as doing a bit of tidying/housekeeping of the pipeline structure).

## Benefits
We will know which stable IDs are duplicated in which species.

## Possible Drawbacks
Pipeline will take longer to run, because the table which we are querying has 800 million rows.

## Testing
Pipeline run successfully with small dataset. Results pending for complete table.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/StableIDs_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/StableIDs_conf.pm
@@ -53,6 +53,15 @@ sub default_options {
   }
 }
 
+sub pipeline_create_commands {
+  my ($self) = @_;
+
+  return [
+    @{$self->SUPER::pipeline_create_commands},
+    'mkdir -p '.$self->o('output_dir')
+  ];
+}
+
 sub resource_classes {
   my ($self) = @_;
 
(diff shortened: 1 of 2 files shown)


james-monkeyshines:
Running the duplicate query across the current stable_id_lookup table produced a report with ~19.5 million duplicate stable IDs, ~5 million of which were genes in core databases. The query took about 8 hours and peak memory usage on the node running it was 3.5GB - not too shabby, given the volume of data.

vinay-ebi (approved):

PR #482, opened 2020-09-25
ENSCORESW-3534: disconnect from database while parsing big files

magaliruffier:

## Requirements

## Description

Implement disconnect_if_idle on ParseSource jobs

## Use case

During the ParseSource analysis, jobs connect to the database at the start of the job then typically process a large file, which can take several minutes. The job is IO bound and does not require the connection to the database any more.
As there can be several thousand parsing jobs when running the pipeline for multiple species, this can create a large number of unused connection to the database, slowing the pipeline down unnecessarily.

## Benefits

Disconnecting from the xref database during the job will reduce contention on the server

## Possible Drawbacks

For short jobs, the disconnection can be more costly.

## Testing

The xref pipeline has been run successfully on multiple species with the implemented change. A decrease in concurrent database connections has been observed.

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/ParseSource.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/ParseSource.pm
@@ -47,6 +47,7 @@ sub run {
             port    => $port,
             user    => $user,
             pass    => $pass });
+  $xref_dbc->disconnect_when_inactive(1);
 
   my $dbi = $self->get_dbi($host, $port, $user, $pass, $dbname);
 


magaliruffier:
Just wondering if anyone had a chance to look at this or whether you need more detail?
This is only for release 104 onwards, so no rush

james-monkeyshines (commented):
The ticket refers to setting the `disconnect_if_idle` property, but the code uses `disconnect_when_active` - I've never quite understood the subtle difference between these two, but many moons ago I was warned off `disconnect_when_active` (I think because it uses up connections faster than they can be released?), and throughout the Production code base we use `disconnect_if_idle`. If you say it's the right thing to do I'll believe you, _naturellement_, but just wanted to double-check this was what you intended?

magaliruffier:
It is a fair question, I have to double-check again what the difference between the two is every time!
The GB eHive cheatsheet has the best explanation I've found so far, which can be summarised as such:
- disconnect_if_idle will disconnect the worker but not the runnable
- disconnect_when_inactive will disconnect the whole runnable
The assumption here is that the job is IO limited (parsing the input file) and does not need the connection until the end of the job to write the results
It is not useful if the job reconnects to the database within one minute, because the connection port will not have been released
Thinking about it some more, maybe it would be safer to use disconnect_if_idle instead, sorry for the trouble

PR #484, opened 2020-09-25
Removed `/nfs/nobackup/` usage

marcoooo:

## Description

/nfs/nobackup will be retired ruding this release. moved everything to /hps/nobackup2/production/ensembl/

## Use case

/nfs/nobackup/ temp dir usage.

## Benefits

No missing mount point when running pipeline. 

## Possible Drawbacks

Unexisting expected directory. 

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/DBCopyPatch_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/DBCopyPatch_conf.pm
@@ -37,7 +37,7 @@ sub default_options {
     'registry'      => '',
     'pipeline_name' => $self->o( 'ENV', 'USER' ) . '_DBCopyPatch_' .
       $self->o('ensembl_release'),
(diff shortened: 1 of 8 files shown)


james-monkeyshines (commented):
I'm not sure that `/nfs/production/ensembl/production` is the best place for these files - that's a more-or-less permanent, backed-up location. There is a `/nfs/production/ensembl/nobackup/production` directory, but `/nfs/production/ensembl/nobackup` seems not to be used by anybody. Generally speaking, the idea of using the `/nfs/production/ensembl` directories, rather than `/nfs/production/panda/ensembl`, seems to have been given up by all the teams - there's barely anything timestamped in the last couple of years.
Since the message about `/nfs/nobackup` being retired was sent in error, we don't need to make any changes right now. The filesystem review, for planning the farm cluster migration, has shown that we scatter files across lots of directories, and I think it'd be sensible to rationalise things - but not piecemeal, so I suggest closing this PR.

marcoooo:
We'll come back to this PR another time. if any.

PR #486, opened 2020-09-28
Base decision-making in Core Stats pipeline on datachecks

james-monkeyshines:
## Description
The core stats pipeline was using the geneset and assembly information in the metadata db to decide when it needed to be run. This does not take into account some scenarios which affect the statistics: for example, if the members of a biotype group change, or if 'karyotype_rank' attribs are added. The easiest way to tackle this is to run the appropriate group of datachecks, and run the pipeline for any species which fails. This is a bit redundant, because we don't necessarily need to re-do density calculations if only a biotype has changed (for example), but the pipeline is not _that_ expensive to run, so we tolerate this for the sake of a simpler pipeline to maintain.

## Use case
We ended up with species with out-of-date statistics because we didn't have sufficient triggers in place to re-calculate them in all scenarios. Although we had a group of datachecks for this, there was no mechanism to run them across all divisions every release.

## Benefits
Statistics are correct.

## Possible Drawbacks
It takes a bit longer to run the datacheck group, compared to the assembly/geneset lookup in the metadata db. But it doesn't take ages, the datacheck queries are fairly simple.

## Testing
No unit tests, but pipeline initialised and successfully run to completion for one division.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
@@ -59,8 +59,6 @@ sub default_options {
     bin_count => '150',
     max_run   => '100',
 
-    skip_metadata_check => 0,
-
     pepstats_binary => 'pepstats',
     pepstats_tmpdir => '/scratch',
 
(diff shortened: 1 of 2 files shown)


vinay-ebi (approved):

marcoooo:
Not directly related to this build, but tests are failing here:
https://travis-ci.org/github/Ensembl/ensembl-production/builds/731016935#L13710 @vinay-ebi Is it not related to your latest dump updates? Mind this test failure may have a big impact on EBEYE search and THOAS

PR #501, opened 2020-10-19
Reduce Sample Data time/memory usage

james-monkeyshines:
## Description
The analysis that derives a 'good' example gene/transcript/location was taking hours and large (sometimes >16GB) amounts of memory. This is excessive - it's not important to get the absolute best example gene, we just need something that is in a decent-looking gene tree, and is preferably in RefSeq/UniProt. Rather than spend time optimising, we can just reduce the number of trees we look at in the first place - currently ~6000 trees are examined. Reducing that by an order of magnitude, by limiting the set of results to 500, cuts down both time and memory usage.

## Benefits
Pipeline runs faster, uses less memory.

## Possible Drawbacks
We do not guarantee that we choose the best possible gene, optimised for all our criteria.

## Testing
Run for a test species - new code took 37 mins and 2GB, rather than 2 hr 29 mins and 9GB. Different examples were chosen, but both were fine as sample entry points into the data.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/SampleData_conf.pm
@@ -158,22 +158,20 @@ sub pipeline_analyses {
                               maximum_gene_length => $self->o('maximum_gene_length'),
                            },
       -hive_capacity   => 50,
-      -batch_size      => 10,
       -flow_into        => {
                              '1' => ['RunDataChecks'],
-                             '-1' => ['GenerateSampleData_25GB'],
+                             '-1' => ['GenerateSampleData_mem'],
                            },
       -rc_name          => 'mem',
     },
     {
-      -logic_name       => 'GenerateSampleData_25GB',
+      -logic_name       => 'GenerateSampleData_mem',
       -module           => 'Bio::EnsEMBL::Production::Pipeline::SampleData::GenerateSampleData',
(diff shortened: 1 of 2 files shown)


james-monkeyshines:
Hello @thibauthourlier, I wanted to get a genebuilder's comments on this change, wasn't sure of the best person to ask, please delegate as you see fit...

thibauthourlier (approved):
Oh yes, I don't think there is a need to look at too many trees. We don't really care what are the entry points as long as they don't look like random 1 exon gene.

One comment, because of the thresholds used we should be fine but one thing we might want to control is the location. Maybe there is already code for that. If we have chromosomes we would prefer the entry point to be on a chromosome rather than a small contig. I have no idea if it is something easily feasable.

james-monkeyshines:
> One comment, because of the thresholds used we should be fine but one thing we might want to control is the location. Maybe there is already code for that. If we have chromosomes we would prefer the entry point to be on a chromosome rather than a small contig. I have no idea if it is something easily feasable.

That's not currently done; it is possible, but it's not a simple change. You'd have to refactor the code that retrieves the set of gene trees, to include joins to the genome_db and dnafrag tables - the latter has a length field that could be used in an 'order by', as a proxy for chromosomal regions. (I think that's beyond the remit of Production, so is left as an exercise for the interested reader...)

thibauthourlier:
I'm not sure I'm interested ;)

vinay-ebi (approved):

PR #502, opened 2020-10-21
Webdatafile dump2020

vinay-ebi:
WebDataFile Pipeline for generation of static files with genomic data for Genome Browser in the new Ensembl web site.
Pipeline divided into several steps :

**bootstrapping step**
This is preliminary step of collecting certain information about the genome (such as the number, names, and lengths of the chromosomes), which is stored in the common_files directory, in a subdirectory with the name of the specific genome. 

**retrieval of genes, transcripts, and contigs step**
The step will create a directory under genes_and_transcripts dir and generate the transcripts.bed (further compressed to transcripts.bb) file. Its format, as a list of columns, is as follows:

Chromosome name
Gene start position
Gene end position
Transcript id (versioned)
Strand (forward/reverse)
CDS start (relative to the beginning of the chromosome)
CDS end (relative to the beginning of the chromosome)
Total number of exons
Exon lengths
Exon start positions (relative to the transcript start)
Transcript start position
Transcript end position
Transcript biotype
Transcript designation ("canonical", "mane_select", or "-")
Gene id (versioned)
Gene name
Gene description
Gene biotype

Contigs data is fetched to generated bed files and indexed with the bedToBigBed tool, which produces bigBed files that are eventually consumed by genome browser:

**generation of %GC data step** 
The Step will require GC_analysis Python module, as well as wigToBigWig and bigWigCat USCS tools.
to compute gc percentage 

**generation of variation data step**
The Step generates vcf data from consequences of variations (have the _incl_consequences part in their name). 
the vcf files is written into one or many bed files with the following columns:
column 1 — chromosome name
column 2 — start position
column 3 — end position
column 4 — consequence of the variation


modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
```diff
@@ -0,0 +1,150 @@
...
+    genome_id  => $self->param('genome_id'),
+    species    => $self->param('species'),
+    version    => $self->param('version'),
+    type       => $self->param('type'),  
+    root_path  => path($self->param('root_path'))
+  };
+  my $dba = Bio::EnsEMBL::Registry->get_DBAdaptor( $self->param('species'), $self->param('group') );
+  my $lookup = Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup->new("genome_data" => $genome_data); 
+   my $slice_adaptor = Bio::EnsEMBL::Registry->get_adaptor( $self->param('species'), $self->param('group'), 'Slice' );
+  my $genome = $lookup->get_genome('1');
```
marcoooo, line 67: Shoudn't it be $self->param('genome_id')?

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
```diff
@@ -0,0 +1,142 @@
...
+  my $genome_data = {
+    dbname     => $self->param('dbname'),
+    gca        => $self->param('gca'),
+    genome_id  => $self->param('genome_id'),
+    species    => $self->param('species'),
+    version    => $self->param('version'),
+    type       => $self->param('type'),  
+    root_path  => path($self->param('root_path'))
+  };
+  my $lookup = Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup->new("genome_data" => $genome_data); #"root_path"=> path("/hps/nobackup2/production/ensembl/vinay/test_webdatafile"));
```
marcoooo, line 66: Avoid letting some comments in code (this can be misleading for next dev who wouldn't know if it has been "forgotten" of commented on purpose.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
```diff
@@ -0,0 +1,142 @@
...
+use base qw/Bio::EnsEMBL::Production::Pipeline::Common::Base/;
+use Path::Tiny qw(path);
+use Carp qw/croak/;
+use JSON qw/decode_json/;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::IndexWig;
+use Capture::Tiny qw/capture/;
+use Path::Tiny qw/path tempfile/;
+
+sub param_defaults {
```
marcoooo, line 42: Not sure this overriden method is useful, since doing nothing more than parent?

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
```diff
@@ -0,0 +1,183 @@
...
+use Carp qw/croak/;
+use JSON qw/decode_json/;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::IndexBed;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::variants::VariantsScaler;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::variants::VariationBedWriter;
+use CoordinateConverter qw(to_zero_based);
+
+
+sub param_defaults {
```
marcoooo, line 45: Same comment as above for overridden method.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
```diff
@@ -0,0 +1,183 @@
...
+    species    => $self->param('species'),
+    version    => $self->param('version'),
+    type       => $self->param('type'),  
+    root_path  => path($self->param('root_path'))
+  };
+  my $dba = Bio::EnsEMBL::Registry->get_DBAdaptor( $self->param('species'), $self->param('group') );
+  my $lookup = Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup->new("genome_data" => $genome_data); 
+  my $genome = $lookup->get_genome('1');
+  my $variation_bed_writer = Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::variants::VariationBedWriter->new(root_path => path($self->param('root_path')));
+  ##my $release = get_release_versions();
```
marcoooo, line 73: Same comment about comments.

marcoooo (changes requested):
You have many files failing the Copyright tests. You should add the default LICENSE on every files when needed.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
vinay-ebi: As analysis step run per species and we are creating Genome object for single species in the analysis i used constant value as 1 to get the genome object for current species . i will use species name which makes more readbility

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
vinay-ebi: sure i will remove the unwanted comments

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
marcoooo: Ok fine.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
vinay-ebi: i will remove it

marcoooo (changes requested):
Could you change the ::lib:: to ::Utils:: Seems to be more like the Ensembl convention for these kind of package.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
vinay-ebi: i will remove the unwanted comments

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/GenomeInfo.pm
```diff
@@ -0,0 +1,76 @@
...
+                       root_path   => $self->param('output_path'),   
+                }, 1 );
+
+	} 
+} ## end sub dump
+
+sub genome_info {
+	my ( $self, $species, $compara ) = @_;
+	my $genome;
+	if ( $compara && $compara ne 'multi' ) {
```
james-monkeyshines, line 64: There's a fair amount of code in this module to do with the `$compara` variable, but it only seems to be used to determine whether ENS_VERSION and EG_VERSION should be passed to the `new` method (which is done when `ne` 'multi', which is equivalent to the division not being 'vertebrates' - that would seem a simpler test?) And why not pass the *_VERSION variables for vertbrates anyway?

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,226 @@
...
+}
+
+sub get_assembly_report {
+
+  my ($self, $genome, $dba) = @_;
+
+  my $target_path = $genome->genome_report_path()->stringify;
+  open(my $fh, ">", $target_path);
+
+  my $statement = $self->prepare_sql($genome->species_id());
```
james-monkeyshines, line 82: $statement is not used anywhere, should that be passed via `-SQL` in the next line?

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,226 @@
...
+  my $query_result = $dba->dbc->sql_helper()->execute( -SQL => $self->prepare_sql(1) ); 
+  $self->write_column_headers($fh);
+  $self->write_report($fh, $query_result);
+}
+
+sub report_to_chrom_lookups{
+
+   my ($self, $genome) = @_;
+   my $genome_report = $genome->get_genome_report();
+   my $write_seqs_out = (exists $ENV{ENS_WRITE_SEQS}) ? $ENV{ENS_WRITE_SEQS} : 1; # its temporary one
```
james-monkeyshines, line 93: I think it's better to have all variables defined in the pipeconfig, and passed as necessary; using environment variables is less robust, because it's easy to be unaware of them, or to forget to defined them.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,226 @@
...
+  $genome->chrom_sizes_path->spew(@chrom_sizes);
+
+
+}
+
+sub generate_genome_summary_bed{
+   my ($self, $genome) = @_;
+   my @bed;
+   my $bin_number = 150;
+   if(exists $ENV{ENS_BIN_NUMBER}) {
```
james-monkeyshines, line 121: See previous comment.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,226 @@
...
+sub write_output {
+  my ($self) = @_;
+  my $species        = $self->param('species');
+  my $group          = $self->param('group');
+  my $current_step   = $self->param('current_step');
+  
+  if ($current_step eq 'bootstrap') {
+        my @steps     = @{$self->param('step')};
+        my %each_flow = (
+             'transcripts' => 1,
```
james-monkeyshines, line 194: It's best to leave flow `1` alone, and start at (at least) `2`. This is because pipelines automatically flow down pipe 1; most of the time that's not an issue, and I don't think it'd cause any problems currently. But it can cause unexpected behaviour, particularly if you have conditional dataflow, so it's best to get in the habit of avoiding it.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,226 @@
...
+
+use strict;
+use warnings;
+
+use Bio::EnsEMBL::Registry;
+use base qw/Bio::EnsEMBL::Production::Pipeline::Common::Base/;
+use Path::Tiny qw(path);
+use Array::Utils qw(intersect);
+use Path::Tiny qw(path);
+use Carp qw/croak/;
```
james-monkeyshines, line 37: I don't think croak is used anywhere so could be deleted. In any case, in a hive module context, it's better to use $self->warning() - that gets logged in the hive db, whereas croaks would go to STDOUT, and effectively be lost if run via beekeeper.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
```diff
@@ -0,0 +1,150 @@
...
+
+use Bio::EnsEMBL::Registry;
+use base qw/Bio::EnsEMBL::Production::Pipeline::Common::Base/;
+use Path::Tiny qw(path);
+use Carp qw/croak/;
+use JSON qw/decode_json/;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::GenomeLookup;
+use Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::IndexBed;
+
+sub param_defaults {
```
james-monkeyshines, line 40: Redundant method definition.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
```diff
@@ -0,0 +1,142 @@
...
+}
+
+sub wig_index{
+
+  my ($self, $genome) = @_;
+  my $gc_path = $genome->gc_path(); 
+  my $chrom_report = $genome->get_chrom_report('sortbyname');
+  my @paths;
+  my $indexer = Bio::EnsEMBL::Production::Pipeline::Webdatafile::lib::IndexWig->new(genome => $genome);
+  print STDERR "Finding GC wig files for ".$genome->genome_id()."\n";
```
james-monkeyshines, line 81: Printing to STDERR isn't particularly useful when the code is run within a pipeline - it disappears into the aether. If it's useful to have this info, better to use $self->warnings(); else could be removed.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
```diff
@@ -0,0 +1,183 @@
...
+
+
+sub get_vcf_files {
+
+  my ($self, $genome) = @_;
+  my $genome_type = $genome->type(); 
+  my $species = $genome->species();
+  my $ensembl_release = $self->param('ENS_VERSION');
+  my $ensembl_genomes_release = $self->param('EG_VERSION');
+  my $base_ftp_dir = '/nfs/production/panda/ensembl/production/ensemblftp/';
```
james-monkeyshines, line 129: Paths shouldn't be hard-coded in modules - needs to go in the pipe config and be passed in.

vinay-ebi: ok i will add to  pipeline config

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileGenerateGC.pm
vinay-ebi: sure i will use warning function

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
vinay-ebi: i will remove it

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: ok

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: sure will follow

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: will add to pipeline config

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: i will reove this line , as species id is always 1 i passed it  prepare_sql(1) 
 $query_result = $dba->dbc->sql_helper()->execute( -SQL => $self->prepare_sql(1) );

james-monkeyshines: `species_id` is only 1 for non-collection databases. For collection dbs it is used to distinguish between the species within a database - I don't understand exactly what all the code is doing, but I'd have thought you would need to account for it taking a range of values.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/GenomeInfo.pm
vinay-ebi: in GenomeFetcher module : 
i am setting 
$self->{info_adaptor}->set_ensembl_genomes_release($eg_version);
and 
$self->{info_adaptor}->set_ensembl_release($ens_version);
 
to get desired genome info for given release

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: this sql statement is to create report file similar to  NCBI assembly report (https://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.27_GRCh38.p12/GCA_000001405.27_GRCh38.p12_assembly_report.txt) as chromosome names in assembly reports from NCBI sometimes differed from chromosome names used at EBI. Andy suggested a SQL query for our core database for generating the same data as in NCBI assembly report . we generated genome_report.txt file has the same format as NCBI assembly report.

vinay-ebi: as you mention species id will differ for collection DB , i will consider this and make changes accordingly

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/GenomeInfo.pm
james-monkeyshines: Yes; but my point is that you have a conditional structure here to decide whether to pass the eg_version and ens_version parameters to the `new` method of GenomeFetcher. That conditional is based on whether the compara parameter is "multi". This is just a proxy for the division being "vertebrates", so a) why not simply test the division in the conditional statement? and b) why only pass the version parameters for non-vertebrates?

vinay-ebi: i followed conditional statement similar to GenomeFetcher.
a)  yes your right we can make it simple conditional statement by checking ENS or EG version but currently we run the pipeline both vert and nonvert species together so i followed same condition as GenomeFetcher 
b) mostly i did testing with vert species so i carried away. I will change the code accordingly

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileContigs.pm
```diff
@@ -119,9 +120,10 @@ sub map_data {
   my $old_version = $slice->coord_system()->version();
 
   my @decoded_segments;
-  my $projection = $slice->project('contig'); 
+  try {
```
james-monkeyshines, line 115: This is rather un-Perlish - an eval (https://users.cs.cf.ac.uk/Dave.Marshall/PERL/node114.html) would do much the same without an external library. I guess you've done this because errors arise if there are no 'contig' regions to project to - but if an error happens to occur for any other reason, execution will still continue, which seems undesirable. Maybe a check for a contig coord_system could be done instead?

vinay-ebi: sure will do in perl way by using eval.  ya your right  exception code will not fail for any other error. i will add eval block for eval { $projection = $slice->project('contig') } ;  
  i will check contig coord_system to.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
```diff
@@ -0,0 +1,174 @@
...
+
+
+sub get_vcf_files {
+
+  my ($self, $genome) = @_;
+  my $genome_type = $genome->type(); 
+  my $species = $genome->species();
+  my $ensembl_release = $self->param('ENS_VERSION');
+  my $ensembl_genomes_release = $self->param('EG_VERSION');
+  my $base_ftp_dir = $self->param('ftp_pub_path');
```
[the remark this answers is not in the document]
vinay-ebi: to my knowledge its same for both vert and non vert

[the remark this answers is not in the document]
marcoooo: OK

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileVariation.pm
marcoooo: It doesn't prevent the default path to be in the "default" method though.

vinay-ebi: removed this line and added to default params in latest commits

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,214 @@
...
+  my $query_result = $dba->dbc->sql_helper()->execute( -SQL => $self->prepare_sql(1) ); 
+  $self->write_column_headers($fh);
+  $self->write_report($fh, $query_result);
+}
+
+sub report_to_chrom_lookups{
+
+   my ($self, $genome) = @_;
+   my $genome_report = $genome->get_genome_report();
+   my $write_seqs_out =  1; # its temporary one
```
james-monkeyshines, line 84: Is this OK for now, or should this be parameterised?

james-monkeyshines (approved):
All my previous comments have been addressed, have added a minor query that should not be considered to be a blocker.
I am still unsure, about the hard-coded use of '1' for the species_id values - I suspect this may not work as intended for collection dbs, but I'm not sure. Probably easiest to proceed with code as it is, without further delay, and then test whether it works for all species in a collection db.

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: will parameterise

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/WebDataFile_conf.pm
```diff
@@ -48,59 +47,57 @@ sub default_options {
         run_all         => 0, #always run every species
         step             => [],  
         genomeinfo_yml    => undef,
```
marcoooo, line 49: Something to remove, right?

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
```diff
@@ -0,0 +1,214 @@
...
+=head1 NAME
+
+ Bio::EnsEMBL::Production::Pipeline::Webdatafile::WebdataFileBootstrap;
+
+=head1 DESCRIPTION
+  Compute bootstrap step for webdatafile dumps
+
+=cut
+
+package Bio::EnsEMBL::Production::Pipeline::Webdatafile::WebdataFileBootstrap;
```
marcoooo, line 27: Package name should be renamed to something else, related to comments from Code Review today.  Thanks!

marcoooo (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/WebDataFile_conf.pm
vinay-ebi: yes will remove it

modules/Bio/EnsEMBL/Production/Pipeline/Webdatafile/WebdataFileBootstrap.pm
vinay-ebi: suggesting WebdataFileGenomeAssemblyInfo ?

james-monkeyshines: Fine, but that makes for a long name; the "WebdataFile" bit at the start is redundant because that in the path already, so maybe drop that?

vinay-ebi: sure will do the same for other modules to

marcoooo (approved):

PR #506, opened 2020-10-23
Update linkOut config to support st1 and st1b

radome:
## Description

`generate_LinkOut.pl` script requires a text file with staging server details. With the introduction of "staging b" servers, having a ready made config file for `st1b` is quite handy.

## Testing


marcoooo (changes requested):
Ok forget the previous message, wasn't fully awake! Still the name of the file could as well be added as a env var to the ~/.bashrc of the user, and reused for script call?

marcoooo (approved):

radome:
> Ok forget the previous message, wasn't fully awake! Still the name of the file could as well be added as a env var to the ~/.bashrc of the user, and reused for script call?

IMO I'd rather not add configuration that's not widely used across SOPs in `.bashrc`.

PR #529, opened 2020-12-08
Feature/stable id pipeline

james-monkeyshines:
## Description
The Stable ID and Gene Autocomplete pipelines are not designed to be excuted for a subset of species; they both delete all existing data before populating the tables. This does not fit with the RR paradigm, where we want to add data to the existing tables, on a per-species basis (cf https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-6231). This could also potentially speed up the processing for the main site, where, for example, bacteria might not change from one release to the next.

Modifying the Gene Autocomplete pipeline was fairly straightforward, an 'incremental' flag was introduced, and the pipeline will use an existing database if that is set - the default is `0`, so the current behaviour is maintained if the variable is not set. The code is only run for core databases, so it does not need to check for analyses that are not displayed - a datacheck ensures that these will not exist (https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableGenes.pm).

The Stable ID pipeline has had more significant modifcations - it was essentially a wrapper around a script in the ensembl repo (https://github.com/Ensembl/ensembl/blob/master/misc-scripts/stable_id_lookup/populate_stable_id_lookup.pl), but creating the database and index was done by the pipeline (even though this could be done by the script). The script can only be run for all species on a given server, but we want to be able to use the functionality of the SpeciesFactory module, to have fine-grained control about execution. The script is quite complicated, given that it is doing a fairly straightforward task of selecting stable IDs from a few tables, and inserting them in another db. So the relevant code from the script was extracted and refactored into a new pipeline module. It was then easy to introduce an 'incremental' flag, working in the same manner as the Gene Autocomplete pipeline.

Since all of the functionality of the populate_stable_id_lookup.pl script is now in this pipeline, it seemed to make sense to bring the SQL definition for the Stable ID pipeline into this repo, in the same manner as the Gene Autocomplete pipeline.

The pipelines are doing very similar things, so I also homogenised the module and parameter names, to make it a bit easier to maintain them in tandem.

## Benefits
Faster execution (in incremental mode), more control over what species are processed, no redundant work.

## Drawbacks
A bit slower when run across all divisions (because we're using the registry/API, rather than going direct to the dbs) - this is mostly down to bacteria. But those rarely get updated, so we can use the new incremental functionality to leave all those alone, and just update the verts and other non-verts, for most releases. (We _could_ engineer it to just use new/updated species, but it's quick enough, for now at least, that the additional development isn't worth it - the RR use case effectively does this anyway, by running for a list of species, derived from the metadata db.)

## Testing
Pipelines run successfully to completion, replicating databases produced by the former versions of the pipelines.


marcoooo (approved):

james-monkeyshines:
@ameya1981 The Stable ID pipeline (which generates the `ensembl_stable_ids` database) has been significantly updated to replace all, rather than just some, of the functionality of the `ensemblm/misc-scripts/stable_id_lookup/populate_stable_id_lookup.pl` script. This was done to accommodate the use case of Rapid Release, where we did not want to build the database from scratch every two weeks, but rather just add/update the stable IDs for new/updated species. Further details in the PR description - please let us know if you have any comments/questions/objections...

ameya1981 (approved):
Approving for the stableid lookup changes. The travis test failures are for search dumps tests - assuming not associated with these changes.

PR #540, opened 2021-01-18
Bau/merge release104

james-monkeyshines:
Merging release branch to master (required by imminent ProteinFeature-pipeline-related PR).

PR #542, opened 2021-01-18
New pipeline for RNA gene xref annotation

james-monkeyshines:
## Description
A new, small pipeline to add RNAcentral checksum-based xrefs to the RNA genes in a core db. This is the analog of UniParc xref functionality which is now in the ProteinFeatures pipeline (#541).

## Use case
It is intended for Rapid Release only (although it could be used for non-vert divisions, because I don't think RNAcentral xrefs are part of their SOP), because GOA provide their annotations against RNAcentral accessions, so we need those to join things up and maximise our GO terms. It is quick to run: loading the checksums takes ~8 mins, and the annotation itself ~5 mins per species.

## Benefits
Annotate RNA genes with a "hub" xref, which can be used to link through to all the individual RNA resources. 
Maximise our ability to use GO annotations provided by the GOA team.

## Possible Drawbacks
It's still another pipeline that we need to run, even if it is simple and quick.

## Testing
Pipeline initialised and run to completion for a subset of RR species. Some of those had RNAcentral xrefs already, and this pipeline annotates almost the same set, a difference of 1 or 2 xrefs. The differences aren't a problem, xrefs can change slightly over time (this pipeline won't replace any xrefs that exist from other sources anyway) - rather, the message is that the new pipeline is getting the expected level of annotation.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/RNAGeneXref_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/RNAGeneXref_conf.pm
@@ -1,0 +1,223 @@
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::Production::Pipeline::PipeConfig::RNAGeneXref_conf;
+
+use strict;
+use warnings;
+
+use base ('Bio::EnsEMBL::Production::Pipeline::PipeConfig::Base_conf');
+
+use Bio::EnsEMBL::Hive::PipeConfig::HiveGeneric_conf;
+use Bio::EnsEMBL::Hive::Version 2.4;
+
+use File::Spec::Functions qw(catdir);
+
+sub default_options {
+  my ($self) = @_;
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProteinFeatures_conf.pm
```diff
@@ -47,7 +47,7 @@ sub default_options {
     max_dirs_per_directory  => $self->o('max_files_per_directory'),
 
     # InterPro settings
-    interproscan_version    => '5.46-81.0',
+    interproscan_version    => '5.48-83.0',
```
marcoooo, line 50: Is is sensible to use some kind of "get whatever version is available by the time you run the pipeline"? 
for instance, with RR, we do:
``
INTERPRO_VERSION=`interproscan.sh -version | egrep version | tr -s ' ' | cut -d ' ' -f 3`
``

Then pass the version to the pipeline, but I guess this could be automated as well. right?

james-monkeyshines: It would be handy to be able to tell the pipeline to use whatever version is available on the farm, but you would also need to determine the source db versions automatically - if that turns out to be not too tricky, I can add it to this PR. If not, I think the SOP for RR will need to change, and we'll have to work out how to get it to use the correct conf file without passing in the version; because the config also contains the versions for the different InterPro sources, passing the interproscan_version parameter would break the new version-checking functionality.

marcoooo: It definitely need more thinking, this is fine for this PR to stay the same. We'll think again about it at some point (More generally speaking about having something reliable to get the current software versions available in our software environment). Wasn't VersionService not an attempt to do that?

james-monkeyshines: Ah, just noticed that this discussion is on the wrong PR - this one is just for the RNA Gene Xref pipeline, the Protein Features one is #541. The interproscan version only appears here because #540 has not been merged yet, once it has, those commits will disappear from this PR.

james-monkeyshines:
Pipeline now documented here: https://www.ebi.ac.uk/seqdb/confluence/display/GTI/RNA+Gene+Xref+pipeline

marcoooo (approved):

PR #554, opened 2021-02-19
use species scientific name instead of display name

vinay-ebi:
New FTP Dump Pipeline creates the directories with species display name from meta table to dump the data.
While executing this pipeline for Rapid release 16 we observed The regex used to replace the comman name enclosed in the parenthesis and gca is failing for few cases example : **Cavia tschudii(Montane guinea pig) - GCA_004027695.1** 
for time being we suggest to use species scientific names instead of species display name

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base.pm
@@ -84,14 +84,15 @@ sub species_name {
   $self->throw("Missing dba parameter: species_name method") unless defined $dba;
 
   my $mca = $dba->get_adaptor("MetaContainer");
-  my $species_name = $mca->single_value_by_key('species.display_name');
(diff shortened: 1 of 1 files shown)


james-monkeyshines (commented):
I don't think this is the correct fix - we do not already use the scientific name because we cannot rely on it having the correct value. The meta value needs to be corrected to include a space before the opening parenthesis, and it will then be correctly parsed/dumped.
We have agreed, along with Genebuild, Web and everyone else on a very specific format for this field - the correct format needs to be used, otherwise downstream processing fails.
This has happened before, and Genebuild should have already changed their pipelines to include the space - if that's not been done, we need to request that.

vinay-ebi:
ok, can we include datacheck in handover to validate the species display name as per agreed format so we wont enconter this problem

james-monkeyshines:
Not easily, because this is a Rapid Release specific constraint - but yes, ultimately that'd be a good idea.

PR #556, opened 2021-03-23
Pipeline mirror load

vinay-ebi:
Copy grch37 mart to mirror-2


modules/Bio/EnsEMBL/Production/Pipeline/MirrorLoad/ListDatabase.pm
```diff
@@ -73,7 +81,7 @@ sub run {
 	}
 
 	#list all functional dbs
-	if( !( $self->param('process_grch37') || $self->param('process_mart') ) || $self->param('run_all') ){
+	if( !( $self->param('process_grch37') || $self->param('process_mart') || $self->param('process_grch37_mart') ) || $self->param('run_all') ){
```
marcoooo, line 84: imho, Don't know the actual impact, but since this is same target for `all` grch37, I would just have remove the mart filter in case of grch37.

marcoooo (commented):
My last comment may not need an actual update in the code, but could be a way to simplify a bit the whole pipeline.

modules/Bio/EnsEMBL/Production/Pipeline/MirrorLoad/ListDatabase.pm
vinay-ebi: added grch37 mart  flag to give more flexibility to copy only functional dbs or mart dbs for grch37

vinay-ebi: added grch37 mart flag to give more flexibility to copy only functional or mart dbs for grch37

marcoooo: Not sure this is a real use case, we'll see with usage :-)

marcoooo (approved):
Would myself fix the travis in this PR, even though not completely related. OR merge other PR to fix the build before merging this one.

PR #560, opened 2021-04-09
Load RNAcentral-derived GO terms in GPAD pipeline

james-monkeyshines:
## Description
The GOA team provide terms annotated against UniProt and RNAcentral identifiers (plus some other misc. sources) - however, a bug was introduced 4 years ago that prevented the loading of RNAcentral-derived terms, with the exception of those linked to microRNAs (which represent a very small amount of the total RNAcentral annotation). 

This fix continues to treat microRNAs differently, loading the precursors rather than the product (because we have the former not the latter in our dbs), but uses the given accession in all other cases.

## Testing
Test pipeline was run for panda - gained ~5000 GO terms linked to RNA genes.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
@@ -166,13 +166,18 @@ sub run {
       } elsif ($db =~ /RNAcentral/) {
         $self->log()->debug("Adding linkage to RNAcentral");
         $is_transcript = 1;
-        # Accession is the version with taxonomy id appended, e.g. URS0000007FBA_9606
-        # We store as URS0000007FBA, so need to remove everything from the _
-        # precursor_rna could be a list of accessions
-        my @precursor_rnas = split(",", $precursor_rna) if $precursor_rna;
-        foreach my $rna (@precursor_rnas) {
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/GPAD/LoadFile.pm
```diff
@@ -166,13 +166,18 @@ sub run {
...
         $self->log()->debug("Adding linkage to RNAcentral");
         $is_transcript = 1;
-        # Accession is the version with taxonomy id appended, e.g. URS0000007FBA_9606
-        # We store as URS0000007FBA, so need to remove everything from the _
-        # precursor_rna could be a list of accessions
-        my @precursor_rnas = split(",", $precursor_rna) if $precursor_rna;
-        foreach my $rna (@precursor_rnas) {
-          $rna =~ s/_[0-9]*//;
-          $db_object_id = $rna;
+        # For miRNA, we want the precursor(s), rather than the db ID,
```
marcoooo, line 169: Sounds sensible, could you just add an example in the comments? for future reference. Thanks!

marcoooo (commented):
Simply add a more detailed comment and that'll be all right.

james-monkeyshines:
> Simply add a more detailed comment and that'll be all right.
Done!

marcoooo (approved):

PR #562, opened 2021-04-15
Split APPRIS/TSL pipeline, because these need to be run independently

james-monkeyshines:
## Description
Split APPRIS/TSL pipeline, because these need to be run independently.
Renamed directories and files for clarity and consistency.
Removed dependency on metadata db, don't need that complexity.

## Use case
APPRIS is dependent on TSL, so we need to annotate TSL first, and subsequently do APPRIS. Until now, this didn't really matter, because TSL has not been updated in a long time, so the pre-existing TSL annotation sufficed. However, human and mouse databases are now created differently, and any existing TSL annotation is deleted. The Havana team thus need to annotate TSL prior to handover, so we need an independent pipeline. The Production team will continue to run the APPRIS pipeline.

## Benefits
All data required by APPRIS is in the database at handover, so there is enough time before release for the APPRIS annotations to be calculated and loaded.

## Possible Drawbacks
None I'm aware of.

## Testing
Both new pipelines run for human, mouse, fruit fly. Replicated results of previous pipeline.


james-monkeyshines:
Further details in ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-6570

luca-drf (approved):

PR #567, opened 2021-05-20
Rewrite Rapid Release Xref TSV file dump to use SQL not API 

james-monkeyshines:
## Description
The Xref dumping is very slow using the API; this rewrites the module to use a handful of SQL queries, which are vastly quicker (~5 mins, rather than hours), and probably easier to understand than the previous set of API calls. Memory usage is reduced too. 

These changes also fix a bug where sources for GO terms were not correctly retrieved - the API-based code was not filtering on the transcript, so the sources were reported for all transcripts. Further, only one 'linkage_type' was reported in these cases; this is a per source property, however, so the value was moved to the compound 'sources' field. And the name of the field changed too, because 'dependent_sources' could be taken to mean that these were annotations dependent on the GO term, rather than the other way around.

## Benefits
File dumping for Rapid Release is much quicker - periodic updates for protein features and GO terms, for all species, are easier within the 2-week timeframe.

## Possible Drawbacks
Less robust against schema changes, but those are very unlikely.

## Testing
Run successfully for all current RR species.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base.pm
@@ -196,6 +196,8 @@ sub has_seq_region_attribs {
 sub assert_executable {
   my ($self, $exe) = @_;
 
+  $exe =~ s/\s+.+//;
+
   if ( !-x $exe ) {
     my $output = `which $exe 2>&1`;
     chomp $output;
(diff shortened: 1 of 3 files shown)


james-monkeyshines:
@luca-drf I've assigned this one to you so that you can appreciate the much better formatting of my commit messages, following your retreat session...

luca-drf (approved):

PR #572, opened 2021-06-08
Codon migrations

vinay-ebi:
Modified pipeline conf file according to new codon server

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/OntologiesLoad/MartLoad.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/OntologiesLoad/MartLoad.pm
@@ -21,6 +21,7 @@ package Bio::EnsEMBL::Production::Pipeline::OntologiesLoad::MartLoad;
 use strict;
 use warnings FATAL => 'all';
 use base ('Bio::EnsEMBL::Production::Pipeline::Common::Base');
+use Path::Tiny qw(path);
 
 sub fetch_input {
 }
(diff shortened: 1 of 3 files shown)


marcoooo (approved):

modules/Bio/EnsEMBL/Production/Pipeline/OntologiesLoad/MartLoad.pm
```diff
@@ -42,7 +42,7 @@ sub write_output {
     $self->run_system_command("$srv -e \"DROP database IF EXISTS $mart\"");
     $self->run_system_command("$srv -e \"CREATE database $mart\"");
     $self->log("Building mart database $mart");
-    my $TMP_SQL = "/scratch/mart.sql";
+    my $TMP_SQL = "/hps/scratch/mart.sql";
```
james-monkeyshines, line 45: TSC say we need to have user names in the path. You can get that in Perl with `$ENV{'USER'}`. However, you then need to create the sub-directory, so it gets a little more complicated... Could either 1) build and create the path in the conf file and pass a parameter in here; or, 2) do something like we do with File::Path and File::Temp in https://github.com/Ensembl/ensembl-production/blob/release/105/modules/Bio/EnsEMBL/Production/Pipeline/ProteinFeatures/InterProScan.pm. (Or, these days, Path::Tiny is my go-to library for filesystem stuff, it has nice wrappers around that functionality.) Note that the latter option gives us an easy way to remove the file when the module exits.

james-monkeyshines (changes requested):

james-monkeyshines (approved):

PR #573, opened 2021-06-17
Modify Protein Features pipeline for Codon cluster

james-monkeyshines:
## Description
On Codon we have a centrally-managed version of InterProScan, which is much easier for us to manage, but needs a little bit of adaptation in the pipeline. (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-6789)

## Use case
For a given instance of the pipeline, we want to run the same version of interproscan; but the interproscan team might update the 'current' symlink mid-pipeline. So, resolve the 'current' link to a distinct version at the start of the pipeline, and pass that version to subsequent modules that need the path to the executable.

A corollary of the above point is that the specific version does not need to be hard-coded in the config file.

Versions of the Interpro source dbs are also no longer hard-coded in the conf, but are instead determined by running interproscan.sh with no parameters, to get a list of the dbs and their versions.

A more general Codon issue is defining where data is stored on the filesystem - some default paths are defined in the Base pipeline config, so that we can make all of our pipelines consistent.

## Benefits
Protein Features pipeline works on Codon.
We never have to upgrade InterProScan again!

## Possible Drawbacks
None I'm aware of.

## Testing
Pipeline succesffuly run to completion for two vertebrates.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/ProteinFeatures/InterProScanVersionCheck.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/ProteinFeatures/InterProScanVersionCheck.pm
@@ -24,23 +24,34 @@ use warnings;
 use base ('Bio::EnsEMBL::Production::Pipeline::Common::Base');
 
 use File::Fetch;
+use File::Spec::Functions qw(catdir);
 use Path::Tiny;
+
+sub param_defaults {
+  my ($self) = @_;
+  
+  return {
+    %{$self->SUPER::param_defaults},
+    interproscan_version => 'current',
+  };
+}
 
 sub run {
   my ($self) = @_;
+  my $interproscan_path    = $self->param_required('interproscan_path');
   my $interproscan_version = $self->param_required('interproscan_version');
-  my $interproscan_exe     = $self->param_required('interproscan_exe');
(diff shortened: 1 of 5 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/ProteinFeatures/InterProScanVersionCheck.pm
```diff
@@ -54,9 +65,29 @@ sub run {
         $self->throw("Could not find version in service file:\n$service_version_file\n$data");
       }
     }
+
+    # To guard against the 'current' version changing during pipeline
```
marcoooo, line 69: I don't see it from the code (surely my fault), but is the process retrieving "where" the symlink is pointing to and register it? Or is it done another way?

james-monkeyshines: No, that would probably have been a cleverer way to do it. Instead, it works on the assumption that the subdir for a specific version has the format `interproscan-<version>`, and since we have just saved the version into a parameter, we just concatentate things together to get the path.

vinay-ebi (approved):

PR #574, opened 2021-06-18
Configure Production pipelines for Codon cluster

james-monkeyshines:
## Description
Update pipelines to use Codon filesystem paths; made parameter names consistent across all Production pipelines.
Also did a bit of tidying, getting rid of unused files.

## Benefits
Pipelines work on Codon! Better management of our use of nobackup and scratch space.

## Possible Drawbacks
Various parameters have changed names, for consistency, so I'd expect the odd teething problem while we update SOPs etc.

## Testing
Minimal - syntax has been checked, but pipelines with updated conf files have not been executed. However, the changes are not drastic, so significant problems are not anticipated.


james-monkeyshines:
Note that the changes here might be clearer if you refer to an earlier commit to the Base_conf: https://github.com/Ensembl/ensembl-production/pull/573/commits/1c08ea6

luca-drf (approved):

PR #576, opened 2021-06-24
GFF3 export, formatted for ENA submission.

james-monkeyshines:
### Generate GFF3 in the format required for ENA submission.
This differs from our normal export in a few ways; the versioned stable IDs for all features are set as 'Name' attributes (only gene is required to have a name, but it doesn't hurt to be consistent).

The versioned INSDC accessions are used to identify sequences; if a region is a chromosome, the name is given by the 'chromosome' attribute. Circular regions have the 'Is_circular' attribute; and if a region needs anything other than the standard Codon table, there is a 'transl_table' attribute.

Some aspects of the file are in line with the NCBI spec: gene_id, transcript_id, protein_id attributes all contain the versioned stable ID; exon order is given by the 'exon_number' attribute rather than 'rank'. The biotype of a gene is given by a 'gene_biotype' attribute, rather than 'biotype'.

Two types of sequence edits are represented in the file.
- Protein-level edits are indicated by setting the 'exception' attribute (from the NCBI spec), with the edits in the format requested by ENA in the 'transl_except' attribute. This format is different from the NCBI spec, so in a belt-and-braces approach, the NCBI-format edits are given in a 'Note attribute'.
- Transcript-level edits are also indicated by setting the 'exception' attribute. Information on transcript-level edits was not requested by ENA, but is provided, in the same format as requested for proteins, in a 'Note' attribute.

The ENA requirements necessitated a bit of additional flexibility in the generic dump files from which we inherit.
- For custom paths, directories are now created if they do not exist (which is the existing behaviour if the paths are to be the standard RR format).
- The general GFF3 export has an extra parameter, to indicate whether the GFF3Serializer should add GFF3 headers and metadata. Usually we want this, but not for the ENA format, because it defaults to using the seq_region names, not INSDC synonyms, and it's hard to override this. Instead, we rely on the 'gt' tidying step to add in the necessary details.

The pipeline to dump the GFF3 files can largely inherit from the generic dumping pipeline; the functionality needed to run pre-dump datachecks, to verify that we have INSDC sequence accessions, is added to the generic pipeline, because that is generally useful (and also makes inheritance more straightforward). The companion datacheck PR is https://github.com/Ensembl/ensembl-datacheck/pull/388.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base_Filetype.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/FileDump/Base_Filetype.pm
@@ -54,6 +54,10 @@ sub fetch_input {
 
   if ($self->param_is_defined('custom_filenames')) {
     $filenames = $self->param('custom_filenames');
+
+    foreach my $file (values %{$filenames}) {
+      $self->create_dir(path($file)->parent);
+    }
   } else {
     my $overwrite = $self->param_required('overwrite');
 
@@ -112,17 +116,7 @@ sub filenames {
     $dir = $self->param('output_dir');
   }
 
-  # If the parent dir does not exist, we will need to
-  # update the group permissions, so that other teams
-  # can write to it once we have created it.
-  my $parent_exists = path($dir)->parent()->exists();
-
-  path($dir)->mkpath();
-  path($dir)->chmod("g+w");
-
-  if (! $parent_exists) {
-    path($dir)->parent()->chmod("g+w");
-  }
+  $self->create_dir($dir);
 
   my %filenames;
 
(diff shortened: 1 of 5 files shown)


marcoooo:
Not directly related I presume, but the build is failing because of this: https://travis-ci.com/github/Ensembl/ensembl-production/builds/230756123#L21764 Any parameter retired recently from CoreStats Pipeline?

james-monkeyshines:
> Not directly related I presume, but the build is failing because of this: https://travis-ci.com/github/Ensembl/ensembl-production/builds/230756123#L21764 Any parameter retired recently from CoreStats Pipeline?

It's because I renamed a path parameter in an update for Codon migration, I've got a fix ready to go.

marcoooo (approved):

PR #577, opened 2021-06-25
Ontology Load pipeline updates

james-monkeyshines:
## Description
When testing the parameterisation of the Ontology Load pipeline on Codon, in order to update the SOP, I noticed that the datachecks are almost guaranteed to fail. The comparison with the previous release will usually throw up something, but will rarely be anything to worry about. If that test is critical, the pipeline will generally fail, and not create the dbcopy jobs.

Split the datacheck execution into advisory and critical sets; the comparison check is made advisory (cf https://github.com/Ensembl/ensembl-datacheck/pull/385), and the report is emailed.
Renamed a couple of variables and reformatted a bit, to be consistent with other pipelines.

## Benefits
Pipeline won't always fail.

## Possible Drawbacks
If there _is_ a significant drop in the terms between releases, then we rely on someone paying attention to emails. An alternative solution would be to make a more complex datacheck, that allowed for a little leeway in the counts - but it's hard to know what is acceptable, and it probably is dependent on the ontology.

## Testing
Test pipeline run successfully on Codon.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/OLSLoad_conf.pm
@@ -38,27 +38,49 @@ sub default_options {
 
     return {
         %{$self->SUPER::default_options},
-        'base_dir'         => $self->o('ENV', 'BASE_DIR'),
-        'srv_cmd'          => undef,
-        'wipe_all'         => 0,
-        'wipe_one'         => 1,
-        'verbosity'        => 2,
-        'ols_load'         => 50,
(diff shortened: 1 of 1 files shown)


marcoooo (approved):

marcoooo:
Validated even with build errors due to updates in pipeline config, right?

james-monkeyshines:
> Validated even with build errors due to updates in pipeline config, right?

Yes, the build errors are fixed over here: #575

james-monkeyshines:
Following comments from the ontology team, it may be that we don't expect counts to drop, and that such failures should halt the pipeline. However, the changes here are still ok to merge; if we don't have any advisory checks, that section of the pipeline will simply not run. But at least we have the option, for future use.

PR #581, opened 2021-07-09
Module and pipeline to update 'website_packed' status for Rapid Release.

james-monkeyshines:
## Description
"Packing" the data for a species, to enable the website display to be quicker, is generally required whenever a database changes. Packing can be a time-consuming process, however, so for Rapid Release, where we have a two-week release cycle and lots of databases that aren't changing from one release to the next, we want to avoid redundant work.

A new field has been added to the metadata db to track the packed status, and there is a script in the metadata repo to update this. But it makes it easier for us to manage if we also have that functionality in a module that we can add to pipelines, and a pipeline that will apply to all species derived from a SpeciesFactory module.


marcoooo (approved):

marcoooo:
@james-monkeyshines  Regarding build errors, the one about the copyright, I don't know which way is the best.... 104 was started in 2020, but ended in 2021.... The one about the search dumps, I believe has been fixed on master by @vinay-ebi a while ago, after branching to 104. I would believe we should backport this bugfix on branch. Any opinion?

james-monkeyshines:
> @james-monkeyshines Regarding build errors, the one about the copyright, I don't know which way is the best.... 104 was started in 2020, but ended in 2021.... The one about the search dumps, I believe has been fixed on master by @vinay-ebi a while ago, after branching to 104. I would believe we should backport this bugfix on branch. Any opinion?

I don't have a strong opinion on the copyright, but I guess we should follow what the 'ensembl' repo does, and the release/104 branch has not been universally updated to 2021 (e.g. https://github.com/Ensembl/ensembl/blob/release/104/modules/Bio/EnsEMBL/AltAlleleGroup.pm). So I don't think there's anything to do there.

As for backporting the bugfix, it wouldn't hurt, but I don't think it is essential - it was caused by a typo in the test, rather than the actual code (https://github.com/Ensembl/ensembl-production/commit/1ad62bd), so nothing was truly broken.

PR #584, opened 2021-07-13
Increase memory for GFF3 ENA format dumps.

james-monkeyshines:
Some jobs were hitting the memlimit during the step where the file is renamed after being sorted. For reasons I couldn't make out, this was not being properly detected and spawning a subsequent 'high_mem' job. But might as well increase the default memory to avoid teh problem in the first place.


marcoooo (approved):

marcoooo (approved):

marcoooo:
I think the same should be opened against release/104 branch, I reckon this is the one I used for dumping RR species.

james-monkeyshines:
> I think the same should be opened against release/104 branch, I reckon this is the one I used for dumping RR species.

Rapid release runs from the latest branch of ensembl-production; the ensrapid user has release/105 checked out.

PR #590, opened 2021-08-04
Feature/db copy pipeline

james-monkeyshines:
## Description
At various points in the release cycle, we want to bulk copy sets of databases from one server to another. Currently, we repeatedly generate lists of databases based on regexes, and submit them to the db copy service in a loop. It's easy to either miss dbs out, or include ones you shouldn't with this approach; and it also does not maximise the efficiency of the db copy service, which is happier dealing with one job handling 200 dbs with a '%core%' regex, rather than 200 separate jobs.

Further details and use cases are in a ticket: [ENSPROD-7032](https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7032)

This PR introduces a new pipeline for submitting databases to the dbcopy service. By default, all species-level databases are copied; marts, compara dbs, and other multi-species dbs can also be copied. Arbitrary database names can also be given.

Each different bulk copy usually involves some ancillary work, like deleting older dbs from the target server, or renaming dbs and applying patches. So this functionality is included in the pipeline: databases can optionally be deleted from the target server,
for either the same release as the dbs, or a specified past release; and the pipeline can increment the release numbers in
database names and apply patches.

A related task is granting read-only permissions on a server - this typically happens at a later stage of the release than the copying, so the functionality cannot be incorporated in the copying pipeline. But, we can have a simple new pipeline to do this, using the same modules that have been developed for the copying pipeline.

SOPs created/updated for all use cases, see the related JIRA ticket for a [list of links](https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7032?focusedCommentId=367986&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-367986).

## Benefits
Copying is more efficient, easier to administer and track, and less error-prone.

## Possible Drawbacks
None I can think of.

## Testing
Test pipelines run for a range of copying/deleting/renaming scenarios.


marcoooo (approved):
Sounds promising!

PR #629, opened 2021-11-30
Ensprod 5803 exon phase

vinay-ebi:
Added exon phase more details in ticket : https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-5803

### DIFF
--- a/modules/Bio/EnsEMBL/Production/DBSQL/BulkFetcher.pm
+++ b/modules/Bio/EnsEMBL/Production/DBSQL/BulkFetcher.pm
@@ -280,6 +280,8 @@ sub get_transcripts {
 		e.seq_region_start as start, 
 		e.seq_region_end as end,
 		e.seq_region_strand as strand,
+		e.phase as phase,
+		e.end_phase as ens_phase,
 		et.rank as rank,
 		'exon' as ensembl_object_type
 		FROM transcript t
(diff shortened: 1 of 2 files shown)


marcoooo (changes requested):
I don't think this will be enough, the fetcher is used in many places, did you check its usage across production repo, and potential impact on some existing other dumps? 
Another thing, I believe is if you meant to make it available in search dumps, that might need a modification in the related `Formatter` class.

vinay-ebi:
> I don't think this will be enough, the fetcher is used in many places, did you check its usage across production repo, and potential impact on some existing other dumps? Another thing, I believe is if you meant to make it available in search dumps, that might need a modification in the related `Formatter` class.

fetcher used in  DumpCore pipeline  alone other than unified dumps. 
Unified dumps pipeline uses GeneFetcher.pm module to fetch the gene information  from Raw sql used in BulkFetcher module if we want to add this exon phase information in formater class GeneFetcher we need to query geneinforamtion again with exon phase info. Previously we added CDS and SO terms in Bulkfetcher similar way. I will check the impact of  DumpCore with this changes

marcoooo (approved):

marcoooo (approved):

marcoooo (changes requested):
Any chance to fix the build on travis prior to merge?

vinay-ebi:
> Any chance to fix the build on travis prior to merge?

Build failed due to :  missing Copyright [2016-2022] EMBL-European Bioinformatics Institute

marcoooo (approved):

PR #632, opened 2021-12-06
seq region name and length added to unified dumps

vinay-ebi:
Sequence region name and length added to unified dumps


.travis.yml
```diff
@@ -29,7 +29,7 @@ before_install:
 - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-test.git
 - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl.git
 - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-hive.git
-- git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-compara.git
+- git clone --branch main --depth 1 https://github.com/Ensembl/ensembl-compara.git
```
marcoooo, line 32: I'd rather apply the same principle as for metadata app instead:
- add a `COMPARA_BRANCH` to exported vars
- set it to `main` then overwrite it with $TRAVIS_BRANCH in the IF statement like `26`

marcoooo (changes requested):

marcoooo (approved):

PR #636, opened 2021-12-08
EP-7421: DBCopy module won't resubmit a copy job is an equivalent one is already running

luca-drf:
## Description

DBCopy module won't resubmit a copy job is an equivalent one is already running

## Benefits

In case the DBCopy hive analysis needs to be restarted, a new copy job won't be submitted if another one is already  running

## Possible Drawbacks

It's crucial to maintain the DBCopy database in a state where no more than one equivalent jobs are marked as not completed

## Testing

- [x] Have you run the entire test suite and no regression was detected?


marcoooo (approved):
Don't know exactly  why file Scraper source were included, but that doesn't hurt.

PR #656, opened 2022-03-04
Updated the OMIM download key

jmgonzmart:

## Requirements

## Description

This PR updates the URL that the Xref pipeline uses for downloading the OMIM data.

## Use case

OMIM requires active accounts for individuals/institutions wishing to download their data. Each account has a download key associated to it that is included in the file URLs. 

We were recently contacted by OMIM because a list of their download keys had been exposed in a public website. This list included the key associated to the Ensembl account, so they urged us to get a new key. The old key has been discontinued.

## Benefits

The Xref pipeline will be able to download the OMIM file again.

## Possible Drawbacks

None.

## Testing

I have tested that the new URL works for downloading the data using 'wget'. I haven't tested the pipeline using the updated URL, but I don't expect any issue.

Dependencies
------------

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       identification within third-party archives.
 
    Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-   Copyright [2016-2021] EMBL-European Bioinformatics Institute
+   Copyright [2016-2022] EMBL-European Bioinformatics Institute
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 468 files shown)


marcoooo (commented):
I am happy to validate it @jmgonzmart but one question, since this key is public in our repo, should we turn this info in a private one?

jmgonzmart:
You are right, it didn't occur to me that this is actually a public repository and may have contributed to that list of keys. If it is possible to keep this info private, it will certainly be the best thing to do.

marcoooo (approved):

marcoooo:
For the record, I discussed the matter with #Infra and we decided to allow this key for 107, then regenerate it for next release while storing it in a safer place.

PR #667, opened 2022-04-25
Features/perl 5.26

marcoooo:
## Description

Added 5.26 to PERL build 

## Use case

Expected upgrade to PERL 5.26

## Benefits

Code testing works with 5.26

## Possible Drawbacks

TBD

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Search/JSONReformatter.pm
+++ b/modules/Bio/EnsEMBL/Production/Search/JSONReformatter.pm
@@ -58,7 +58,6 @@ sub process_json_file {
   {
     local $/ = '}';
     while (<$fh>) {
-      s/\\\\//g;	    
       my $obj = $json->incr_parse($_);
       if ( defined $obj ) {
         $n++;
--- a/modules/Bio/EnsEMBL/Production/Search/JSONReformatter.pm
+++ b/modules/Bio/EnsEMBL/Production/Search/JSONReformatter.pm
@@ -58,6 +58,7 @@ sub process_json_file {
   {
     local $/ = '}';
     while (<$fh>) {
+      s/\\\\//g;	    
       my $obj = $json->incr_parse($_);
       if ( defined $obj ) {
         $n++;
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,10 @@ matrix:
       perl:
         - "5.14"
         - "5.26"
-    - language: python
-      python:
-        - "3.7.6"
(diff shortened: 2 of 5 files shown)


vinay-ebi (approved):

vinay-ebi:
All checks passed in PR: https://github.com/Ensembl/ensembl-production/pull/709 this mite be a transient error

modules/Bio/EnsEMBL/Production/Search/JSONReformatter.pm
```diff
@@ -58,6 +58,7 @@ sub process_json_file {
   {
     local $/ = '}';
     while (<$fh>) {
+      s/\\\\//g;
```
marcoooo, line 61: I am not sure I understand well your comment, is the issue related to test content? with a hidden char inside? 
Since those test files are issued from test sample DB, is it not safer to fix the sample data instead?

vinay-ebi: There is an unexpected escape character '\' in dictionary  { "hidden_remark":"formerly known as \\", "auth…}, which causes the failure, Not sure why test data differs from the Perl version. ya, it's safer to edit the test data.

marcoooo: Cheers! let's do that then. Thanks!

marcoooo: but I believe we can keep this clean up, just in case some data are containing this nasty char (I think the data are issued from actual db fetched to generated the test content), so it might be best to keep the change, and fix the original DB. to stay on the safest side.

vinay-ebi: sure we will fix the actual data

marcoooo: Can you point me to the test file? so we can act on the Test/Live DB to fix this?

vinay-ebi: t/41-searchdumps_reformat_advsearch.t

marcoooo: Well, that's the test itself, not the related test datasource. :-)

vinay-ebi (approved):

PR #682, opened 2022-06-06
default pub ftp paths changed as per codon

vinay-ebi:
default ftp path updated for assembly converter script


marcoooo (approved):
Don't forget to update the SOP to be run from a data mover queue.

PR #695, opened 2022-07-14
Fixes for non-GIFTS species

sgiorgetti:
## Description

Changes for fixing:
- duplicated `protein_feature` records in core DBs
- missing new `protein_feature` records for non-GIFTS species
- ".A" hit names for `protein_feature` 

## Use case

Create AlphaFold protein features for additional species, which are not necessarily represented in GIFTS DB.

## Benefits

Fixing bugs and mitigating current pipeline design issues.

## Possible Drawbacks

For non-GIFTS species, the Uniprot entry is retrieved from Xref data. These Xref are required to be of `info_type` DEPENDENT, but that is not necessarily the case for some non-vertebrates species.
In these latter cases, the AlphaFold data cannot be mapped, `protein_feature` are not created, and data checks will fail.

These drawbacks will be addressed in `release/109`.

## Testing

The changes were manually successfully tested. No automatic testing procedure was developed.

Dependencies
------------

No new deps.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/AlphaFold/HiveLoadAlphaFoldDBProteinFeatures.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/AlphaFold/HiveLoadAlphaFoldDBProteinFeatures.pm
@@ -82,6 +82,7 @@ sub param_defaults {
       cs_version => undef,
       species => undef,
       rest_server => undef,
+      debug => undef,
     }
 }
 
@@ -110,6 +111,10 @@ sub fetch_input {
 
   $self->hrdb_set_con($core_dba,"core");
 
+  print "Cleaning up old protein features and analysis\n" if defined($self->param('debug'));
+  $self->cleanup_protein_features('alphafold_import');
+
+  print "Initiating MakeAlphaFoldDBProteinFeatures and creating the analysis object\n" if defined($self->param('debug'));
   my $runnable = Bio::EnsEMBL::Production::Pipeline::AlphaFold::MakeAlphaFoldDBProteinFeatures->new(
     -analysis => new Bio::EnsEMBL::Analysis(-logic_name => 'alphafold_import',
                                             -db => 'alphafold',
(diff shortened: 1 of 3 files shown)


sgiorgetti:
The `alphafoldUpdate.py` scripts has been fixed to cope with the `protein_feature.hit_name` = ".A" issue only.
A new version of the script is going to be made available separately to improve performance.

modules/Bio/EnsEMBL/Production/Pipeline/AlphaFold/HiveLoadAlphaFoldDBProteinFeatures.pm
```diff
@@ -110,6 +111,10 @@ sub fetch_input {
 
   $self->hrdb_set_con($core_dba,"core");
 
+  print "Cleaning up old protein features and analysis\n" if defined($self->param('debug'));
```
marcoooo, line 114: is print going to log in Hive log table?

sgiorgetti: Not sure: that's why the IF with the "debug" parameter, which is by default undefined.
I used it on STDOUT with `standaloneJob.pl`.
I wanted to double-check with Arne when he's back next week.

If you agree, I can fix it for 109. If you don't, I'll remove the "print" altogether.

marcoooo: I believe this can be change to log instead:
```use Bio::EnsEMBL::Utils::Exception qw(throw info);```
Then use 
```info(sprintf("No data found for species %s in GIFTS DB using endpoint %s and assembly %s. Message:\n%s", $self->{'species'}, $self->{'rest_server'}, $self->{'cs_version'}, $@));```

Issued from past Arne PR :-)

marcoooo: If too complicated, just comment the lines, and we'll make the change another time.

sgiorgetti: Just pushed changes. Tested OK for "candida albicans" only for the sake of being quick.

PR #705, opened 2022-08-15
fix for recurring error

vinay-ebi:
Datacheck output creating duplicate jobs in the speciesfactory with resulted running jobs multiple times and failing to meet desired inputs.
 details are in ticket : https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7770


marcoooo (changes requested):
You may open it back to 107 as well, because this is the current branch used for RR. 
Thanks

marcoooo (approved):

PR #707, opened 2022-08-17
New param  forced_species added to compute the corestats for  given s…

vinay-ebi:
New param  forced_species added to compute the corestats for  given species irrespective of datacheck results.
more details in ticket : https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7983

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
@@ -159,7 +159,7 @@ sub pipeline_analyses {
       -hive_capacity   => 50,
       -batch_size      => 10,
       -flow_into       => WHEN(
-                            '#species# ~~ @{#forced_species#}' || '#datachecks_failed#'=>
+                            '#species# ~~ @{#forced_species#} || #datachecks_failed#'=>
                             [
                               'CodingDensity',
                               'PseudogeneDensity',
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
```diff
@@ -157,7 +159,7 @@ sub pipeline_analyses {
       -hive_capacity   => 50,
       -batch_size      => 10,
       -flow_into       => WHEN(
-                            '#datachecks_failed#' =>
+                            '#species# ~~ @{#forced_species#} || #datachecks_failed#'=>
```
marcoooo, line 162: Is this doing a diff between #species# and #forces_species#? if so what happens if we have `run_all` and no species specified?

vinay-ebi: its checks the  species exists in the forced_species array , here species is type string generated from from previous analysis step `speciesfactory` . if run_all and no species specified species_factory generates the species names from which each species compared in the forced_species list .
image

marcoooo: ok so the new param `run_all_forced` would override this to prevent us to set the whole species list in `forced_species`, right? I believe this could be evaluated differently, I mean you don't need a specific parameter, just test if `run_all` is set?

vinay-ebi: added new param run_all_forced without using existing run_all as considering the  use-case if we want to select all species with run_all and force few species to compute corestats even though dc failed.  

run_all_forced param will compute corestats for all species without considering data check failures and forced_species list.

wanted to keep these two params separate.

marcoooo (approved):

PR #709, opened 2022-09-01
added flag to include or exclude readthrough gene count

vinay-ebi:
New flag added to include or exclude readthrough gene count 
More info  in jira ticket : https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7999

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/CoreStatistics_conf.pm
@@ -63,7 +63,6 @@ sub default_options {
 
     history_file => undef,
     forced_species => [],
-    include_readthrough => 0,
     exclude_species_readthrough => ['homo_sapiens', 'mus_musculus', ],
     run_all_forced => 0,
   };
@@ -218,7 +217,6 @@ sub pipeline_analyses {
       -max_retry_count => 1,
       -hive_capacity   => 50,
       -parameters      => {
-                            include_readthrough => $self->o('include_readthrough'),
 			    exclude_species_readthrough => $self->o('exclude_species_readthrough'),
                           },
       -flow_into       => {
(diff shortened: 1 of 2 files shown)


marcoooo:
@thibauthourlier @andrewyatz if you could have a look, we are entering a domain where I'd rather have a second/third advice!

jmgonzmart (changes requested):
It looks fine to me except for the following. In GeneCount.pm l. 49, the call to the _get_alt_attrib_codes_ method allows for a second parameter ($include_readthrough), but when the method is defined (l. 135), this second parameter is not included.

jmgonzmart (commented):
If I understand correctly the new block of code in GeneCount.pm, the **$has_readthrough** and **$include_readthrough** flags will be set to zero for human and mouse by default, so the codes for readthrough gene counts won't be added to the list of attribute codes. Those gene counts are required too.
Secondly, the default gene counts (eg. 'coding_cnt'), shouldn't include readthrough genes. For instance, the current number of coding genes in release 107 ('coding_cnt') is 19804. This excludes readthrough genes. However, the code in the get_feature_count method doesn't exclude readthrough genes. Are the current values in the database coming from this pipeline or is there a subtraction of readthrough genes behind the scenes? Should the GeneCount module address this directly?

vinay-ebi:
> If I understand correctly the new block of code in GeneCount.pm, the **$has_readthrough** and **$include_readthrough** flags will be set to zero for human and mouse by default, so the codes for readthrough gene counts won't be added to the list of attribute codes. Those gene counts are required too. Secondly, the default gene counts (eg. 'coding_cnt'), shouldn't include readthrough genes. For instance, the current number of coding genes in release 107 ('coding_cnt') is 19804. This excludes readthrough genes. However, the code in the get_feature_count method doesn't exclude readthrough genes. Are the current values in the database coming from this pipeline or is there a subtraction of readthrough genes behind the scenes? Should the GeneCount module address this directly?

We included the readthrough attributes required.  The `count_all_by_Slice` return the gene count including the readthrough gene, we subtracted the readthrough genes coding_rcnt from the overall gene count coding_cnt. In the previous release, we manually removed the readthrough genes count.

jmgonzmart:
Thanks for the clarification. The code makes sense now. It would just need to subtract the number of coding genes with readthrough transcripts from the total coding gene count in the alt regions too, since this was also done in release 107 (coding_acnt = coding_acnt - coding_racnt).

marcoooo (approved):

PR #725, opened 2022-10-25
Xref download changes

TamaraNaboulsi:

## Requirements

## Description

Recently, some xref source files have been corrupted or incomplete which is leading to job failures. This prompted us to re-think about the way we download these source files.

## Use case

The download logic has been changed:
- ftp have been changed to http/https as support has been largely stopped and http/https is more stable
- related to that, the way we download RefSeq files have changed (to read file names from a catalog to support name pattern)
- files existing on our local ftp (ftp.ebi.ac.uk) are copied directly from there using datamover
- added better error handling regardless of download method, as currently jobs still succeed even if got incomplete files

This change also includes handling a recent change in UniProt files which now include versioned stable IDs in the sprot Ensembl mapping lines. This is affecting isoform xrefs which seem to mainly affect humans, but some cases might occur for vertebrates.

## Benefits

- Better error handling to catch corrupted files early in the pipeline
- Not using deprecated ftp

## Possible Drawbacks

None.

Possible future improvements:
- comparing file checksums to ensure downloaded files are valid and complete
- look into quicker http/https download methods (other than HTTP::Tiny or LWP::Simple)

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
@@ -35,6 +35,8 @@ use JSON;
 use Bio::EnsEMBL::Hive::Utils::URL;
 use Bio::EnsEMBL::DBSQL::DBConnection;
 use Text::Glob qw( match_glob );
+use File::Copy qw(copy);
+use LWP::Simple;
 use Bio::EnsEMBL::Utils::IO qw/slurp/;
 use Bio::EnsEMBL::Utils::URI qw(parse_uri);
 
@@ -63,10 +65,19 @@ sub create_db {
 }
 
 sub download_file {
-  my ($self, $file, $base_path, $source_name, $db, $skip_download_if_file_present, $release) = @_;
+  my ($self, $file, $base_path, $source_name, $db, $extra_args) = @_;
 
+  # Create uri object and get scheme
   my $uri = URI->new($file);
   if (!defined $uri->scheme) { return $file; }
+
+  # Get extra parameters
+  my $skip_download_if_file_present = $extra_args->{'skip_download_if_file_present'} || 0;
+  my $release = $extra_args->{'release'} || undef;
+  my $rel_number = $extra_args->{'rel_number'} || undef;
+  my $catalog = $extra_args->{'catalog'} || undef;
+
+  # Create file download path
   my $file_path;
   $source_name =~ s/\///g;
   my $dest_dir = catdir($base_path, $source_name);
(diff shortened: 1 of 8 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
```diff
@@ -63,18 +65,84 @@ sub create_db {
...
     $dest_dir = catdir($base_path, 'Checksum');
   }
   make_path($dest_dir);
-  if ($uri->scheme eq 'ftp') {
+
+  # If file is in local ftp, copy from there
+  if ($file =~ /ftp.ebi.ac.uk/) {
+    # Construct local path
+    my $local_file = $file;
+    $local_file =~ s|http://ftp.ebi.ac.uk/pub/|/nfs/ftp/public/|;
```
marcoooo, line 93: Should it match the `https` protocol instead?

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/gencode_sources.json
```diff
@@ -16,7 +16,7 @@
     {
       "name" : "UniParc",
       "parser" : "ChecksumParser",
-      "file" : "ftp://ftp.ebi.ac.uk/pub/contrib/uniparc/upidump.lis",
+      "file" : "http://ftp.ebi.ac.uk/pub/contrib/uniparc/upidump.lis",
```
marcoooo, line 19: Should it be `https` instead? I tested this uri, and it's available with this protocol

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/gencode_sources.json
```diff
@@ -105,42 +105,50 @@
...
       "name" : "RefSeq_dna",
       "parser" : "RefSeqGPFFParser",
-      "file" : "ftp://ftp.ncbi.nih.gov/refseq/H_sapiens/mRNA_Prot/*rna.gbff.gz",
+      "file" : "https://ftp.ncbi.nih.gov/refseq/H_sapiens/mRNA_Prot/*rna.gbff.gz",
       "method" : "--bestn 5",
       "query_cutoff" : 90,
       "target_cutoff" : 90,
-      "release" : "ftp://ftp.ncbi.nih.gov/refseq/release/release-notes/RefSeq-release*.txt",
-      "priority" : 2
+      "release" : "https://ftp.ncbi.nih.gov/refseq/release/release-notes/RefSeq-release*.txt",
```
marcoooo, line 112: Just a remark, I believe the code is actually using the FTP protocol to download the file.

marcoooo (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/Base.pm
TamaraNaboulsi: Hi Marc. Here we copy the files from our local ftp location rather than downloading, which was part of the change.

modules/Bio/EnsEMBL/Production/Pipeline/Xrefs/gencode_sources.json
TamaraNaboulsi: Yes the code uses ftp protocol if the link is an ftp one. However, since ftp is no longer supported, we've decided to change this to https (which is available) for more stability.

marcoooo: What do you mean exactly by `supported` ? 
image

marcoooo (approved):

PR #726, opened 2022-11-04
Added timestamp directory syncing

marcoooo:
## Description

Added analysis to enable syncing timestamped directories to public FTP

## Use case

Rapid Release FTP dumps

## Benefits

Symlinks enabled.

## Possible Drawbacks

Might not need to be done, and better to actually rsync with `-L` parameter to resolve the link when publishing. Need discussion

## Testing

Dependencies
------------


marcoooo:
Timestamped directory won't be synced. But we should check that no symlinks stay around after Pipeline completion. Only README.txt should be symlinks.

PR #747, opened 2023-01-04
Checksum.pl updates to remove `ensgen` user test

marcoooo:
## Description

`ensgen` user has been deprecated, so we don't need anymore this requirement, though now the default ftp user is `ensemblftp`, so changed that. The lit of autorised user is extended to our usual virtual users. 

## Use case

Checksums.pl manual run

## Benefits

No need any more to add the `-ignore_user` when running with ensemblftp.

## Possible Drawbacks

None

## Testing

Dependencies
------------

### DIFF
--- a/scripts/ftp/checksums.pl
+++ b/scripts/ftp/checksums.pl
@@ -17,8 +17,9 @@
 use strict;
 use warnings;
 
-my $DEFAULT_USER = 'ensgen';
-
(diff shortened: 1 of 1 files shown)


vinay-ebi (approved):
Build failed due to a copyright issue will be fixed in subsequent PR

PR #761, opened 2023-01-11
fix Blast file names incorrectly after repeat masker rerun

vinay-ebi:
## Description

FTP core dump pipeline creates new blast index files only if the species has a new/updated/renamed genome else it copies from the previous release directory. If repeat_masker is rerun without any updates on the genome the pipeline will copy the blast files from the previous release with the old repeat masker date. which breaks the blast search results. 

## Use case

Blast files are named correctly after repeat masker rerun.

## Benefits

Check if blast files are named with the correct repeat masker name.

## Possible Drawbacks

Access the core DB meta for repeat masker date for each species

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/FASTA/CopyNCBIBlastDNA.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/FASTA/CopyNCBIBlastDNA.pm
@@ -106,12 +106,28 @@ sub get_blast_files {
   my $base = $self->param('ftp_dir');
   my $type = $self->param('type');
   my $old_path = File::Spec->catdir($base, "release-$old_release", 'ncbi_blast', $type);
+  my $repeatmask_value = $self->get_repeatmask_value($self->param('species'));
   my $filter = sub {
     my ($filename) = @_;
-    return ($filename =~ /$species\./) ? 1 : 0;
+    return ($filename =~ /$species\./ && $filename =~ /$repeatmask_value\.dna/) ? 1 : 0;
   };
   my $files = $self->find_files($old_path, $filter);
+  if(! scalar @$files){
+    $self->throw("No Blast index files found in previous release for species $species with repeatmasker date $repeatmask_value");
+  }
   return $files;
+}
+
+sub get_repeatmask_value {
+  my ($self, $species) = @_;
+  my $core_adaptor = Bio::EnsEMBL::Registry->get_DBAdaptor($species, 'core');
+  my $res_array = $core_adaptor->dbc()->sql_helper()->execute_simple( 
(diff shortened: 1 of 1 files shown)


scripts/misc/check_tmux.sh
```diff
@@ -1,4 +1,20 @@
 #!/bin/bash
+#!/usr/bin/env perl
```
marcoooo, line 2: Well spotted for the licence missing, but I guess this line should be removed. 
```suggestion
```

marcoooo (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/FASTA/CopyNCBIBlastDNA.pm
```diff
@@ -106,12 +106,28 @@ sub get_blast_files {
...
 }
 
+sub get_repeatmask_value {
+  my ($self, $species) = @_;
+  my $core_adaptor = Bio::EnsEMBL::Registry->get_DBAdaptor($species, 'core');
+  my $res_array = $core_adaptor->dbc()->sql_helper()->execute_simple( 
+   	  -SQL => qq/select max(date_format( created, "%Y%m%d")) from analysis a join meta m on (a.logic_name = lower(m.meta_value)) where meta_key =?/, 
+	  -USE_HASHREFS => 1, 
+	  -PARAMS => ['repeat.analysis']);
+  print($res_array->[0]);
```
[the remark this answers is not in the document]
vinay-ebi: Ya its not needed , Its for debug and forgot to remove

modules/Bio/EnsEMBL/Production/Pipeline/FASTA/CopyNCBIBlastDNA.pm
```diff
@@ -106,12 +106,27 @@ sub get_blast_files {
   my $base = $self->param('ftp_dir');
   my $type = $self->param('type');
   my $old_path = File::Spec->catdir($base, "release-$old_release", 'ncbi_blast', $type);
+  my $repeatmask_value = $self->get_repeatmask_value($self->param('species'));
   my $filter = sub {
     my ($filename) = @_;
-    return ($filename =~ /$species\./) ? 1 : 0;
+    return ($filename =~ /$species\./ && $filename =~ /$repeatmask_value\.dna/) ? 1 : 0;
```
marcoooo, line 112: BTW shouldn't it be a `or` statement instead?

vinay-ebi: file name matches both species names and repeat masker date : 
ex: Homo_sapiens.GRCh38.20141216.dna
species name : Homo_sapiens and repeatmasker date: 20141216

marcoooo: ok fine! Thanks

marcoooo (approved):

PR #762, opened 2023-01-12
Feature/2020datafile genomeinfos

vinay-ebi:
## Description

2020 Datafile Perl script has a dependency on the genome info yml file which contains species genome info.

## Benefits
This method fetches genome info for species from ensemble metadata, no need to update the yml config file with genome info for new species and easy integration into the NetFlow pipeline for 2020species

## Possible Drawbacks

Need to concise all production common python methods, which can be used for future Nextflow pipeline 

Dependencies
------------
metadata database


vinay-ebi:
move this method to metadata-api

PR #763, opened 2023-01-25
Bugfix/early dumps

vinay-ebi:
## Description

FTP dumping pipeline on the updated species and generate the following dumps  GTF,  GFF3,  FASTA

## Use case
Dump the files as per new FTP dumps layout for vert and nonverts into **ftp_dumps** dir

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/EarlyDumps/MetadataReport.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/EarlyDumps/MetadataReport.pm
@@ -113,8 +113,8 @@ sub species_wise_metadata_files {
 	my @non_vert_combine_files = ();
 	my $cmd_str = "";
 	my $division_specific_param = {    'ftp_dir'          => $self->param('ens_ftp_dir'),
-                                           'base_path'        => catdir($self->param('base_path'),  "/release-".$self->param('ens_version')."/ftp_dumps"), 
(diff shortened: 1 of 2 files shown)


marcoooo (approved):

modules/Bio/EnsEMBL/Production/Pipeline/EarlyDumps/MetadataReport.pm
```diff
@@ -106,11 +106,15 @@ sub species_wise_metadata_files {
...
+	
+	if(@{$self->param('division')}){
+	  @division_list = @{$self->param('division')};
+	}
 	my @non_vert_combine_files = ();
 	my $cmd_str = "";
 	my $division_specific_param = {    'ftp_dir'          => $self->param('ens_ftp_dir'),
-                                           'base_path'        => catdir($self->param('base_path'),  "/release-".$self->param('ens_version')), 
-                                           'dump_dir'         => catdir($self->param('base_path'),  "/release-".$self->param('ens_version')),
+                                           'base_path'        => catdir($self->param('base_path'),  "/release-".$self->param('ens_version')."/ftp_dumps"),
```
marcoooo, line 116: For these lines, may be we should just change the SOP, so the base_path is actually the base_path, and not forcing the subpath structure

marcoooo (changes requested):
May be we should only change the

marcoooo (approved):

PR #767, opened 2023-02-13
Release/110

EbiArnie:
Changes are:
- The store() method on the adaptors returns undef (instead of throw()ing). Handle this with die()
- flock should wait until the lock is acquired. But it can be interrupted, e.g. by a signal. Usually the runnable would go ahead and try to lock the LevelDB. That would be locked, the runnable would error out. But to have a nicer and more precise error message, handle flock errors with die()
- The unlocking of the Perl lock file happened before the unlock of the DB. This could lead to other workers getting errors trying to lock the LevelDB. Now the LevelDB is always unlocked first.

Tested on five species on Codon.


marcoooo (approved):

marcoooo:
Thanks @EbiArnie for this quick fix, is it ready to go with another run? Or should I expect another PR regarding Collection db processing?

PR #771, opened 2023-02-26
Feature/checksum pipeline

vinay-ebi:
**
## Description
New pipeline to generate secure hash  using sha512 and md5 for genomic sequences (toplevel, cda, cdna and peptide) and add the generated hash value into respective attribute tables

## Benefits
Generate secure hash for individual species and sequence type

More details added into the ticket : https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-8593


vinay-ebi:
wrong base

PR #775, opened 2023-03-02
Read data for GIFTS from file, not REST endpoint

EbiArnie:
## Description

So far, we used the REST API endpoint to retrieve GIFTS data. This turned out to be error prone. Fetching data this way takes many hours. If the process is interrupted, the error would be ignored and Uniparc data would be used. Now we use dump files instead. They take seconds to create and to ingest.

## Use case

To actually have data based on GIFTS inserted.

## Benefits

Takes seconds instead of hours. Less error prone.

## Possible Drawbacks

Currently, GIFTS DB dumps have to be created manually. Code and HOWTO exist in production directory.

Tested against 5 of the 6 species currently in GIFTS.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/AlphaFold/InsertProteinFeatures.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/AlphaFold/InsertProteinFeatures.pm
@@ -163,17 +163,19 @@ sub run {
     my $mappings;
 
     if (defined $self->param('cs_version')) {
-        # insert the Ensembl-PDB links into the protein_feature table in the core database
-        info(sprintf("Calling GIFTS endpoint for species %s using endpoint %s and assembly %s.\n", $species, $self->param('rest_server'), $self->param('cs_version')));
-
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/AlphaDBImport_conf.pm
```diff
@@ -72,6 +72,9 @@ sub default_options {
         user         => 'ensadmin',
         alphafold_data_file => '/nfs/ftp/public/databases/alphafold/accession_ids.csv',
         uniparc_data_file => '/nfs/ftp/public/databases/uniprot/current_release/knowledgebase/idmapping/idmapping_selected.tab.gz',
+        # Do not blindly increase the version here. GIFTS data is seldom
+        # refreshed - the current version probably has no GIFTS data.
+        gifts_dir    => '/nfs/production/flicek/ensembl/production/alphafold_gifts/107',
```
marcoooo, line 77: I would prefer this path to be a parameter, with no default, and we set it in the SOP. so we'll know when to change it/override it. and this way no way to launch the pipeline with older data than expected because we forgot the parameter. Or may just be the "gift_version" to be mandatory, if you prefer.

marcoooo (changes requested):
do we have a script to generate the alphafold files somewhere? would be interesting to have it in our repo as well, so we can run it in our SOP when new gift version is released.

PR #777, opened 2023-03-15
renamed rr_version to rr_ens_version

vinay-ebi:


marcoooo (changes requested):
BTW, may be good to change compara_host_uri as well to "undef" so it becomes mandatory to run the pipeline. right?

vinay-ebi:
> BTW, may be good to change compara_host_uri as well to "undef" so it becomes mandatory to run the pipeline. right?

I think that will be a problem when we run only ftp dumps , we force to pass this param

marcoooo (approved):

PR #782, opened 2023-04-03
Update LoadFile.pm

dpopleton:

## Requirements

## Description
A simple fix in the logic. This would return an error rather than checking for defined as you can't call an undefined function like that in perl. It should have an identical functionality, except this will work.

## Use case

GPAD loadFile when you have translation without transcripts

## Benefits

The pipeline will no longer bug out when loading plants.

## Possible Drawbacks

## Testing

Dependencies
------------


marcoooo (commented):
I believe somehow we expect the job to fail, because if we have a `tgt_protein` set it `must` come from a translation. Am I right thinking that?

marcoooo (approved):

PR #801, opened 2023-06-07
Update new service url

vinay-ebi:
set new service url for dbcopy api


marcoooo (changes requested):
This value is present in multiple files https://github.com/search?q=repo%3AEnsembl%2Fensembl-production%20services.ensembl&type=code Could you update all of them?

marcoooo (approved):

PR #803, opened 2023-06-08
Added a failure if gzip fails to complete properly

dpopleton:

## Requirements

## Description

Modified the TSV's to output errors when a problem with gzip is encountered.

## Use case

If Gzip fails, the original file will remain and it will show an error.

## Benefits

No more hidden gzip problems.

## Possible Drawbacks

The pipeline will fail more. It was failing already, but we will see it.

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/TSV/DumpFileXref.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/TSV/DumpFileXref.pm
@@ -130,7 +130,7 @@ sub _write_tsv {
     my $unzip_out_file = $out_file;
     `gzip -n $unzip_out_file`;
 
-    if (-e $unzip_out_file) { `rm $unzip_out_file`; }
+    if (-e $unzip_out_file) { die "gzip failed"; }
   } else {
     # If we have no xrefs, delete the file (which will just have a header).
     unlink $out_file  or die "failed to delete $out_file!";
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/TSV/DumpFileXref.pm
```diff
@@ -130,7 +130,7 @@ sub _write_tsv {
     my $unzip_out_file = $out_file;
     `gzip -n $unzip_out_file`;
 
-    if (-e $unzip_out_file) { `rm $unzip_out_file`; }
+    if (-e $unzip_out_file) { die "gzip failed"; }
```
marcoooo, line 133: May be, we should still delete the file? then `die` ?

marcoooo: Just checking, is the `-e` not testing for the file to exists, then it will die if successful?

dpopleton: If the file exists, then gzip failed.
Gzips standard usage
$gzip file.txt
$ls
file.txt.gz

marcoooo: Ah ok, my mistake! I misread: you were testing the actual zip file presence. All good then,

marcoooo (approved):

marcoooo:
You may backport this fix on 110 branch as well, because when rerunning, you'll need to use the `enseven` environment. Right?

PR #818, opened 2023-07-27
Update ChecksumGenerator.pm

vinay-ebi:
Existing delete SQL query considers meta_key `species.db_name`  which exits for collection db alone and fails to delete attributes for non-collection dbs.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Ga4ghChecksum/ChecksumGenerator.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Ga4ghChecksum/ChecksumGenerator.pm
@@ -340,8 +340,7 @@ sub delete_checksums {
                     inner join coord_system using (coord_system_id)
                     inner join meta using (species_id)
                     inner join attrib_type using (attrib_type_id)
-                    where meta_key = 'species.db_name'
-                        and meta_value = ?
+                    where coord_system.species_id = ?
                         and code = ?
             ";
         } elsif ($seq_type eq 'cdna' or $seq_type eq 'cds') {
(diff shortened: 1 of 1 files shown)


andrewyatz (approved):
Makes more sense to me

modules/Bio/EnsEMBL/Production/Pipeline/Ga4ghChecksum/ChecksumGenerator.pm
```diff
@@ -379,7 +376,7 @@ sub delete_checksums {
             my $at_code = $hash_type . "_" . $seq_type;
 
             my $sth = $dba->dbc->db_handle->prepare($sql);
-            my $records = $sth->execute($species, $at_code);
+            my $records = $sth->execute($dba->species_id(), $at_code);
```
EbiArnie, line 379: I'm not sure this is correct. This must match the species that is passed as a parameter.
Since the DBA is fetched just with type "group", not by species, this could be wrong. I don't know exactly how the adaptors are created for multispecies DBs - does it pick the first one?
Anyway - if you are sure this works for multispecies - OK; if not, this should be tested / confirmed first.

EbiArnie (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/Ga4ghChecksum/ChecksumGenerator.pm
andrewyatz: The registry creates an adaptor for every species within a multispecices DB. The `$dba->species_id()` must match the species you're working with otherwise the whole thing falls apart. 

See https://github.com/Ensembl/ensembl/blob/main/modules/Bio/EnsEMBL/DBSQL/BaseAdaptor.pm#L135 which uses a DBAdaptor instance, and injects the species_id into the local adaptor instance upon construction. The DBAdaptor being used is the core database adaptor. 

I think this is right.

I'm more worried why we're hitting the db_handle directly vs. using the `prepare` method on [DBConnection](https://www.ensembl.org/info/docs/Doxygen/core-api/classBio_1_1EnsEMBL_1_1DBSQL_1_1DBConnection.html#af6be0ea3fdb7341b8dd22571f84a8b19)

EbiArnie (approved):
OK

PR #824, opened 2023-09-22
Features/genome metadata hive

dpopleton:
This contains the hive pipeline for loading of metadata into the ensembl_genome_metadata database

## Use case
Loading a full server of databases into the new metadata database.

## Benefits

The main benefit is to use the species factory and the other hive pipeline benefits on loading databases.

## Possible Drawbacks

None

## Testing

Well tested for both unit tests and practical testing.


src/python/ensembl/production/hive/ensembl_genome_metadata/MetadataUpdaterHiveRnaseq.py
```diff
@@ -13,7 +13,7 @@
 
 
 from ensembl.production.hive.BaseProdRunnable import BaseProdRunnable
-from ensembl.production.metadata.updater import CoreMetaUpdater
+# from ensembl.production.metadata.updater import CoreMetaUpdater
```
marcoooo, line 16:
```suggestion
```
If it's not really temporary, best to delete the line altogether

src/python/ensembl/production/hive/ensembl_genome_metadata/MetadataUpdaterHiveOtherfeatures.py
```diff
@@ -13,7 +13,7 @@
 
 
 from ensembl.production.hive.BaseProdRunnable import BaseProdRunnable
-from ensembl.production.metadata.updater import CoreMetaUpdater
+# from ensembl.production.metadata.updater import CoreMetaUpdater
```
marcoooo, line 16:
```suggestion
```
If it's not really temporary, best to delete the line altogether

src/python/ensembl/production/hive/ensembl_genome_metadata/MetadataUpdaterHiveCdna.py
```diff
@@ -13,7 +13,7 @@
 
 
 from ensembl.production.hive.BaseProdRunnable import BaseProdRunnable
-from ensembl.production.metadata.updater import CoreMetaUpdater
+# from ensembl.production.metadata.updater import CoreMetaUpdater
```
marcoooo, line 16:
```suggestion
```
If it's not really temporary, best to delete the line altogether

modules/Bio/EnsEMBL/Production/Pipeline/Metadata/PayloadGenerator.pm
```diff
@@ -40,6 +40,7 @@ sub run {
     my $database_uri = 'mysql://'.$user.':'.$password.'@'.$host.':'.$port.'/'.$dbname;
     my $email = $self->param_required('email');
     my $metadata_uri = $self->param_required('metadata_uri');
+    my $taxonomy_uri = $self->param_required('taxonomy_uri');
```
marcoooo, line 43: As mentioned during our meeting, we do not plan to add any new parameter to the meta data updater main request/ This parameter should be a global parameter set when initializing the pipeline

marcoooo (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/Metadata/PayloadGenerator.pm
dpopleton: Resolved. It is now a global paramater

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/BasePython_conf.pm
```diff
@@ -43,6 +43,17 @@ sub default_options {
     }
 }
 
+sub beekeeper_extra_cmdline_options {
```
marcoooo, line 46: Why do we need an extra BaseConf here? 

I wonder because we have quite the same code for the "Perl" one... which happens to be ready for slurm. This one doesn't seem to be.

src/python/ensembl/production/hive/ensembl_genome_metadata/MetadataUpdaterHiveProcessDb.py
```diff
@@ -22,14 +22,17 @@ class MetadataUpdaterHiveProcessDb(BaseProdRunnable):
     def fetch_input(self):
         try:
             payload = json.loads(self.param_required("payload"))
-            self.param("comment", payload["input"]["comment"])
-            self.param("database_uri", payload["input"]["database_uri"])
+
+            self.param("comment", payload["input"].get("comment", None))
+            self.param("database_uri", payload["input"].get("database_uri", None))
```
marcoooo, line 27: i'd rather have those put in required params, right? We can't load metadata information without it, right?
so I believe this should be something like:
```suggestion
            self.param_required("database_uri", payload["input"].get("database_uri"))
```

marcoooo (changes requested):

src/python/ensembl/production/hive/ensembl_genome_metadata/MetadataUpdaterHiveProcessDb.py
dpopleton: param_required doesn't work that way. It just checks. See line 28. The python code does not have the full functionality of the perl.

marcoooo: I've a doubt, sorry: https://github.com/Ensembl/ensembl-hive/blob/2558e67ec603f0fe706a22143d1f953baa001d7b/wrappers/python3/eHive/process.py#L234

marcoooo: Or, are you actually setting the parameter there, while retrieving it from the payload?

modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/BasePython_conf.pm
dpopleton: Solved by adding the python code to the pipeline itself and swapping to the base_conf

marcoooo (approved):

marcoooo (changes requested):
If you removed the dep to BasePython_conf. can you remove it from the repo then?

dpopleton:
Not merging to main as it is a MVP process

PR #836, opened 2023-10-18
Changed the path for `pipeline_dir` to match our hps simplified direc…

marcoooo:
…tory structure.

## Description

Pipeline dir was set with the extra from sub_dir, removed this part for the pipeline files to be generated under single 'vep_dir' group. which will append the division automatically
Related SOP updated https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Data+dumping#Datadumping-VEPfiles 

## Use case

VEP dumps runs for 112 release. 

## Benefits

No more messy directory structure (the `pipeline_dir` is the `vep_dir`)

## Possible Drawbacks

None

## Testing

Dependencies
------------


vinay-ebi (approved):

nakib103 (commented):
@marcoooo ,

As you said this is not necessary. We just need to keep in mind that the pipeline will originally dump under -
`vep_dir/sub_dir/<division>/dumps`
I have updated the confluence [page](https://www.ebi.ac.uk/seqdb/confluence/display/EV/VEP+dumps#VEPdumps-Directoriesusedinproductionwrappermodule) - the page was based on this commit.

PR #849, opened 2023-11-16
Update/mvp checksum

dpopleton:
## Description
Adds the option to transfer top level checksums to the new metadata database to existing checksum pipeline
Includes an additional pipeline for loading data onto a new pipe

## Use case

In one case this allows us to load a whole server worth of cores checksums into metadata and on the otherhand it allows us to do it as part of our routine.

## Benefits

checksums are now in metadata.

## Testing

Dependencies
------------

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
@@ -33,7 +33,13 @@ sub run {
     my $host = $core_dbc->host;
     my $port = $core_dbc->port;
     my $dbname = $core_dbc->dbname;
-    my $database_uri = 'mysql://' . $user . ':' . $password . '@' . $host . ':' . $port . '/' . $dbname;
+    my $database_uri = '';
+    if (defined($password)) {
+        $database_uri = 'mysql://' . $user . ':' . $password . '@' . $host . ':' . $port . '/' . $dbname;
+    }
+    else {
+        $database_uri = 'mysql://' . $user . '@' . $host . ':' . $port . '/' . $dbname;
+    }
     if ($self->param('populate_mvp') eq "1") {
         $self->dataflow_output_id({ database_uri => $database_uri }, 3);
     }
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
@@ -40,8 +40,7 @@ sub run {
     else {
         $database_uri = 'mysql://' . $user . '@' . $host . ':' . $port . '/' . $dbname;
     }
-    $self->dataflow({ database_uri => $database_uri });
-
+    $self->dataflow_output_id({ database_uri => $database_uri }, 2);
 }
 1;
 
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
@@ -1,0 +1,42 @@
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2023] EMBL-European Bioinformatics Institute
+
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
```diff
@@ -0,0 +1,42 @@
...
+    my ($self) = @_;
+    my $species = $self->param_required('species');
+    my $core_adaptor = Bio::EnsEMBL::Registry->get_DBAdaptor($species, 'core');
+    my $core_dbc = $core_adaptor->dbc;
+    my $user = $core_dbc->user;
+    my $password = $core_dbc->password;
+    my $host = $core_dbc->host;
+    my $port = $core_dbc->port;
+    my $dbname = $core_dbc->dbname;
+    my $database_uri = 'mysql://' . $user . ':' . $password . '@' . $host . ':' . $port . '/' . $dbname;
```
marcoooo, line 36: 1. Not sure it would work with a `ensro` user without password. May be worth a check.
2. Should it be more generic a module? Because we might want to plus any `core` adaptor a a SQLAlchemy uri.

src/python/ensembl/production/hive/ensembl_genome_metadata/ChecksumTransfer.py
```diff
@@ -0,0 +1,124 @@
...
+
+                    # Add to updates list
+                    if update_data:
+                        update_data['assembly_sequence_id'] = assembly_seq.assembly_sequence_id
+                        updates.append(update_data)
+
+            # Perform the bulk update
+            if updates:
+                session.bulk_update_mappings(AssemblySequence, updates)
+                session.commit()
```
marcoooo, line 124: The `with` statement finish **always** with a commit or rollback. I am not sure you need to run the commit:
https://github.com/Ensembl/ensembl-py/blob/4aaafb5a12a658f3bfb08eaae593e639d66811f5/src/python/ensembl/database/dbconnection.py#L185

marcoooo (changes requested):
Need to address the session part

modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
dpopleton: Fixed

modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
```diff
@@ -0,0 +1,48 @@
...
+    my $port = $core_dbc->port;
+    my $dbname = $core_dbc->dbname;
+    my $database_uri = '';
+    if (defined($password)) {
+        $database_uri = 'mysql://' . $user . ':' . $password . '@' . $host . ':' . $port . '/' . $dbname;
+    }
+    else {
+        $database_uri = 'mysql://' . $user . '@' . $host . ':' . $port . '/' . $dbname;
+    }
+    if ($self->param('populate_mvp') eq "1") {
```
marcoooo, line 43: Sorry I missed this one before, it doesn't hurt to feed the dataflow, without any supplementary param to set. Afterwards, if subsequent module wants it, it can use it.

src/python/ensembl/production/hive/ensembl_genome_metadata/ChecksumTransfer.py
```diff
@@ -0,0 +1,123 @@
...
+                        update_data['sha512t24u'] = checksums['sha512t24u']
+
+                    # Add to updates list
+                    if update_data:
+                        update_data['assembly_sequence_id'] = assembly_seq.assembly_sequence_id
+                        updates.append(update_data)
+
+            # Perform the bulk update
+            if updates:
+                session.bulk_update_mappings(AssemblySequence, updates)
```
marcoooo, line 123: We should add a extra carriage return at the end of the file.

marcoooo (changes requested):

modules/Bio/EnsEMBL/Production/Pipeline/Checksum/CreateURI.pm
dpopleton: Fixed to be more generic

marcoooo (approved):

PR #852, opened 2023-11-21
Added gzip module and converted FileDump_conf.pm to perl gzip module

dpopleton:
## Description

This updates the RR filedump pipeline to use the perl gzip tool, rather than the command line gzip as the command line one has been shown to have issues with partial compression.

## Benefits

Faster and more reliable

## Possible Drawbacks

None

## Testing

Tested with a single dump. Fully successful

Dependencies
------------


marcoooo (approved):

marcoooo (changes requested):
I suspect the target branch is not the right one, should be 110 (for RR) and eventually if necessary `feature/MVP`

marcoooo (approved):

marcoooo (approved):

marcoooo:
@dpopleton Not related directly to this PR, but the failures in the build should be explored, seems like the test are run with MySQL 8.0, which is far away from our target one. ...5.7

PR #865, opened 2024-01-24
Xref Fix and Process Xref Slurm resources

dpopleton:
## Description

This PR fixes the failures in the Plant Xref pipeline and provide slurm support for that pipeline.

## Use case

The plants xref have db problems and fail to complete this should fix it. In addition the slurm resource classes were used over the stnandrd Infra resources

## Possible Drawbacks

It could break at some point

## Testing

- [ X] Have you run the entire test suite and no regression was detected?
- [ X] TravisCI passed on your branch


TamaraNaboulsi (approved):
Change in Xrefs/ScheduleSource.pm looks good to me. This change will help solve the issues happening with running the 'ls' command on different codon nodes where it sometimes behaves in a weird way.

PR #866, opened 2024-01-24
Check that pepstats is there

MatBarba:
Fix a bug where the runnable continues running and finishes silently even if the binary for pepstats is not available.

## Description

Add a simple check for when the pepstats binary is not available.

## Use case

Used in Core stats pipeline. Currently if pepstats is missing, PepStats still completes silently, then PepStats_Datachecks finds that no peptide attributes have been added to the database.

## Benefits

Prevent the runnable from silently finishing when it should fail and be restarted with a working binary path.

## Possible Drawbacks

None

## Testing

No tests added. Used this commit to run a core stats pipeline when pepstats was missing in the environment. After adding a working path to a pepstats binary and restarting the PepStats jobs, the next datachecks pass.

Dependencies
------------

None

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/Production/PepStats.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/Production/PepStats.pm
@@ -103,8 +103,9 @@ sub run_pepstats {
   my $attribs = {};
   my $tid;
   close(OUT);
-  foreach my $line (@lines) {
+  die("pepstats command not found ($PEPSTATS)") if ($lines[0] =~ /command not found/);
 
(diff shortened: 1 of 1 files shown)


marcoooo (commented):
Shouldn't we open this fix on 112 / main instead of release branch?

MatBarba:
> Shouldn't we open this fix on 112 / main instead of/and release branch?

Ok I'll have to create a new branch then

marcoooo (approved):

marcoooo (commented):
@MatBarba Should I expect another PR to 112 or main then? Thanks!

PR #867, opened 2024-01-25
Fixed array bug

dpopleton:
Simple Bug fix to properly get memory specification. 
Tested.


marcoooo (commented):
Why 32 is 100 in the end?

marcoooo (approved):

marcoooo (approved):

PR #868, opened 2024-01-29
updates to new genome factory

vinay-ebi:
GenomeFactory 
Fetch genome by dataset  status 
updates the dataset status
batch limit set with default value 50

### DIFF
--- a/src/python/ensembl/production/factories/metadata/genomeFactory.py
+++ b/src/python/ensembl/production/factories/metadata/genomeFactory.py
@@ -12,11 +12,12 @@
 #  limitations under the License.
 
 '''
-Fetch Genome Info from the new metadata api  
+Fetch Genome Info from the new metadata api
 '''
 import json
 import argparse
 import logging
+import re
 import graphene
 from graphene_sqlalchemy import SQLAlchemyObjectType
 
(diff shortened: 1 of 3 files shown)


vinay-ebi:
Failed checks are related to yearly copyright  and could be  ignored for this PR

src/python/ensembl/production/factories/metadata/genomeFactory.py
```diff
@@ -140,13 +161,13 @@ def resolve_genome_list(self, info, filters=None):
                 query = query.filter(OrganismGroup.name.in_(filters.organism_group))
 
             if filters.organism_name:
-                ensembl_names = set(filters.organism_name) - set(filters.anti_organism_name)
-                if ensembl_names:
-                    query = query.filter(Genome.organism.has(Organism.ensembl_name.in_(ensembl_names)))
+                biosample_id = set(filters.organism_name) - set(filters.anti_organism_name)
```
marcoooo, line 164: All this is to be able to filter by which values. I ask because initially the ensembl_name was the production_name, then it became cumbersome to generate those for GBuilders. So we moved away to find something which is representing something real. But I wonder if this is the implemented/intended beheviour here. e.g. if this is to force to process one particular DB, then the test would be done against genome.production_name instead.

src/python/ensembl/production/factories/metadata/genomeFactory.py
```diff
@@ -203,12 +230,14 @@ def get_genomes(
...
+        batch_size: graphene.Int = '50',
+        dataset_status: graphene.String = ['Submitted'],
+        update_dataset_status: graphene.String = None, 
         query_param: graphene.String = None,
 
 ):
     schema = graphene.Schema(query=Query)
-    DATABASE_URI = metadata_db_uri
-    engine = create_engine(DATABASE_URI)
+    engine = create_engine(metadata_db_uri)
```
marcoooo, line 240: We do have some database connections stuff in ensembl-py. Why not using it?

vinay-ebi: Yes, we could use that, will change accordingly

src/python/ensembl/production/factories/metadata/genomeFactory.py
vinay-ebi: Yes, you are right we want to filter the genome uuids concerning the production name &  will also include for filter for biosample id. thanks for spotting this

marcoooo (approved):

vinay-ebi:
params -species , -division, run_all added to be in sync with old production modules

PR #871, opened 2024-02-01
Rework alphafold pipeline cleanup logic

EbiArnie:
- Pieline ran cleanup too early, after the first datacheck finished. Fixed.
- Removed a few newlines that we were storing but don't need


marcoooo (commented):
Is it something coming from a current discussion about current run (We just started the run for 112 so it's a bit too late for this branch, unless this is a bug fix?) If this is not absolutely mandatory for 112 run, please reopen against `main` branch. Thank You!

dpopleton (approved):

EbiArnie:
Yes, this is a bugfix for the current 112 run.

PR #901, opened 2024-03-21
ChecksumGenerator_conf.pm added support for genomefactory

dpopleton:
## Description

Incoporates the new genome factory into the checksum pipeline.

## Use case

The checksum pipeline can be initialized against datasets which are in the submitted stage.

## Benefits

Can be used in our new set up.

## Possible Drawbacks

Has not been tested in an old environment set up and may fail there.

## Testing

Created a test dataset and the factory successfully processed the data.

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ChecksumGenerator_conf.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ChecksumGenerator_conf.pm
@@ -45,6 +45,8 @@ sub default_options {
         'species'        => [],
         'antispecies'    => [],
         'division'       => [],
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ChecksumGenerator_conf.pm
```diff
@@ -45,13 +45,21 @@ sub default_options {
         'species'        => [],
         'antispecies'    => [],
         'division'       => [],
+        'run_all'        => 0,
```
marcoooo, line 48: Why removed those?

marcoooo (approved):

PR #913, opened 2024-04-17
Remove old metadata dependency from Search dump pipeline and add geno…

vinay-ebi:
## Description

Search Dumps Pipeline for new site 

## Use case

MVP processing (thoas dumps)

## Benefits

No dependency on old metadata
Genome Factory in place 

## Possible Drawbacks

None


modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/EnsemblSearchDumps_conf.pm
```diff
@@ -0,0 +1,150 @@
...
+      -analysis_capacity => 10
+    },
+  ]
+}
+
+sub resource_classes {
+  my ($self) = @_;
+  return {
+    %{$self->SUPER::resource_classes},
+    '100GB' => {'LSF' => '-q '.$self->o('production_queue').' -M 100000 -R "rusage[mem=100000]"'},
```
marcoooo, line 146: In new world, I believe we can avoid defining new confs for LSF.

marcoooo (approved):
@dpopleton can you validate the 10000m (could we set it to 100G instead?)

PR #915, opened 2024-04-22
Fixed gencode_primary string

sgiorgetti:
## Description

"GENCODE Primary" tag should be `gencode_primary` also for consistency with E!112

## Use case

Patching the string to comply with the agreed standard: `gencode_primary`

## Benefits

Keep the format consistent since day 1

## Possible Drawbacks

None

## Testing

N/A - patched the string only.

Dependencies
------------

None


marcoooo (approved):
@vinay-ebi I'll leave you the merge

PR #925, opened 2024-05-15
moved ensembl/xrefs to ensembl/production/xrefs

vinay-ebi:
## Description

resolved namespace xref clash between ensembl-py (ensembl/xrefs) with ensembl-production (ensembl/xrefs)
moved ensembl/xrefs to ensembl/production/xrefs

updated xref config file with profile slurm 

## Use case

Xref Reference Download Pipeline
## Benefits

Run on slurm 

## Possible Drawbacks

None

## Testing

Dependencies
------------


TamaraNaboulsi:
Hi @vinay-ebi. Could you please explain why the namespace change is necessary? In my environment, as long as I add both ensembl-py and ensembl-production to my PYTHONPATH, no clash happens and the code is able to find the appropriate modules. Are you getting an error or is it just preferable not to have the paths be the same name?
In any case, if this needs to go ahead then some extra changes need to be added since each module imports `ensembl.xrefs.Base` so this will need to change to `ensembl.production.xrefs.Base`.

dpopleton (approved):

vinay-ebi:
We moved to new modenv environment where its no longer set the PYTHONPATH it enables the virtual env. 
 when we install any of the application ensembl-production or ensembl-py xrefs module will be overwritten by another.

ensembl.xrefs.Base is changed to ensembl.production.xrefs.Base

TamaraNaboulsi (approved):

PR #929, opened 2024-05-23
Updated JSON remodeler to stop Experimental push on scalar is now for…

dpopleton:
This simple fix just stops the perl error: Experimental push on scalar is now forbidden when running the filedumps with new perl version

### DIFF
--- a/modules/Bio/EnsEMBL/Production/Pipeline/JSON/JsonRemodeller.pm
+++ b/modules/Bio/EnsEMBL/Production/Pipeline/JSON/JsonRemodeller.pm
@@ -292,7 +292,10 @@ sub merge_xrefs {
         $obj->{$dbname} = [];
       }
       for my $ann ( @{ $subobj->{$dbname} } ) {
-        push $obj->{$dbname}, $self->copy_hash($ann);
+        if (ref($obj->{$dbname}) ne 'ARRAY') {
+          $obj->{$dbname} = [];
+        }
+        push @{ $obj->{$dbname} }, $self->copy_hash($ann);
       }
     }
   }


modules/Bio/EnsEMBL/Production/Pipeline/JSON/JsonRemodeller.pm
```diff
@@ -292,7 +292,10 @@ sub merge_xrefs {
         $obj->{$dbname} = [];
       }
       for my $ann ( @{ $subobj->{$dbname} } ) {
-        push $obj->{$dbname}, $self->copy_hash($ann);
+        if (ref($obj->{$dbname}) ne 'ARRAY') {
```
EbiArnie, line 295: The code change itself is fine. The new code does in a proper way what the old code did.

I don't know if this line should allow to overwrite a previous value or if it should just initialise to an arrayref.
In the latter case, the if block is unnecessary.
Code like this: `push (@$d, 1);`
will work just fine with a $d that is undef.
The code on line 298 would work without warnings on its own.

The old code would not overwrite, which is probably a safer approach here.
Therefore - I recommend to remove the if block.


I have additional remarks though.

The code and logic in this sub is really not ideal, in my opinion.
If we find a key in $self->{xrefs} or $self->{protein_features}, we will assign a hashref to $obj->{$dbname}:
$obj->{$dbname}->{$key}++;

If we find a key $self->{annotations}, we now assign (and maybe overwrite) with an arrayref:
push @{ $obj->{$dbname} }, $self->copy_hash($ann);

Is that what the rest of the code expects?

It also seems rather convoluted. I have the feeling that we should probably use "exists" to check for the presence of hash keys instead of defined.
It would probably be a better design to always initialise the values to a ref to an empty array / hash when assigning the key.


It probably is, since you didn't change anything else and I assume this worked before.
But from a design perspective, this should be refactored.

So, overall - this change is fine by me with a wish for future improvement.

EbiArnie (changes requested):

PR #948, opened 2024-08-08
Update travis dist to focal

vinay-ebi:
Travis CI is retiring the default distribution 'xenial' by the end of the year and moving to 'focal'. This PR updates the configuration to reflect this change.


dpopleton:
Don't change the resource conf this way. It is better to specify because the default value of M could change in a later update.

vinay-ebi:
> Don't change the resource conf this way. It is better to specify because the default value of M could change in a later update.

Ignore the base conf file this is mainly to test the Travis dist

PR #968, opened 2024-10-21
Add copyFiles pipeline

bilalebi:
## JIRA Ticket
https://www.ebi.ac.uk/panda/jira/browse/EA-1276

## Description

This pipeline copies files from a source directory to a destination using `rsync`. It optionally sends email notifications on completion or failure.

## Use case & Benefits

With this pipeline, we will be able to automatically copy files from one directory to another

```
nextflow run copyFiles.nf -c ../filescopy.config --source /path/to/source/data  --destination /path/to/destination --file_format .vcf --send_email yes --email_recipient user@example.com
```

#### Features

- **File Transfer**: Uses `rsync` to copy files from the specified source to the destination.
- **File Filtering**: Only processes files matching the specified format (e.g., `.vcf`).
- **Optional Email Notifications**: Sends an email with the pipeline's status, controlled by a `send_email` flag.
- **Directory Checks**: Ensures source and destination directories exist before execution.

## Testing

I tested the pipeline locally and it's working as expected
- [ ] Test it on codon

### DIFF
--- a/nextflow/config/filescopy.config
+++ b/nextflow/config/filescopy.config
@@ -1,0 +1,11 @@
+includeConfig './base.config'
+
+params {
+    source          = '/your/source/directory'
+    destination     = '/your/destination/directory'
+    file_format     = 'vcf'
+
+    email_recipient = 'bilal@ebi.ac.uk'
+    // Set this to 'no' if you don't want to send emails
+    send_email      = 'yes'
+}
--- a/nextflow/config/filescopy.config
+++ b/nextflow/config/filescopy.config
@@ -1,4 +1,4 @@
-includeConfig './base.config'
+includeConfig './production.config'
 
 params {
     source          = '/your/source/directory'
--- a/nextflow/workflows/copyFiles.nf
+++ b/nextflow/workflows/copyFiles.nf
@@ -72,6 +72,7 @@ nextflow run copyFiles.nf -c ../filescopy.config --source /path/to/source/data -
 }
 
 process COPY {
+    label 'mem2GB_DM'
     input:
     val src_file
     val dest_dir
--- a/nextflow/workflows/copyFiles.nf
+++ b/nextflow/workflows/copyFiles.nf
@@ -77,7 +77,7 @@ process COPY {
     val dest_dir
 
     """
-    rsync -avz $src_file $dest_dir
+    rsync -avz  --no-o --no-g --no-perms  $src_file $dest_dir
     """
 }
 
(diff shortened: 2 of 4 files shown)


nextflow/workflows/copyFiles.nf
```diff
@@ -0,0 +1,105 @@
...
+""".stripIndent()
+}
+
+process COPY {
+    input:
+    val src_file
+    val dest_dir
+
+    """
+    rsync -avz $src_file $dest_dir
```
dpopleton, line 80:
```suggestion
    rsync -avz  --no-o --no-g --no-perms  $src_file $dest_dir
```

nextflow/workflows/copyFiles.nf
```diff
@@ -0,0 +1,105 @@
...
+  --email_recipient         Email address to receive pipeline status notifications
+                            Ex: user@example.com
+
+Example:
+nextflow run copyFiles.nf -c ../filescopy.config --source /path/to/source/data --destination /path/to/destination --file_format .vcf --send_email yes --email_recipient user@example.com
+""".stripIndent()
+}
+
+process COPY {
+    input:
```
dpopleton, line 75:
```suggestion
    label 'mem2GB_DM'
    input:
```

nextflow/config/filescopy.config
```diff
@@ -0,0 +1,11 @@
+includeConfig './base.config'
```
dpopleton, line 1: We need the slurm configuration from production.config

dpopleton (changes requested):
It is fine, baring the comments, but not too useful at the moment.
I am approving, but we need some sort of DC, even a dummy, before I want to include it in pipelines. I am not sure how you want to do that with a full directory sync, as opposed to single files.
I thought something like this:
`process InitialDatacheck {
    label 'mem2GB_DM'
    input:
    path initial_file

    output:
    path initial_file

    script:
    """
    echo "Running datacheck on initial file ${initial_file}"
    ensembl-datacheck --file ${initial_file} --test=${params.file_type}
    """
}`
But that can come later, if need be.

nextflow/workflows/copyFiles.nf
```diff
@@ -0,0 +1,105 @@
...
+
+// Helper function to check if a directory exists
+def checkIfExists(path, label) {
+    if (!file(path).exists()) {
+        error "${label} directory doesn't exist: ${path}"
+    }
+}
+
+// Function to build and send email
+def sendPipelineStatusEmail(String pipelineName, String status, String recipient) {
```
nwillhoft, line 22: Could we simplify this by using the built-in Nextflow workflow notification report with `-N`? e.g. https://www.nextflow.io/docs/latest/notifications.html#workflow-notification

bilalebi: The reason why I though of creating a custom email is because `-N` doesn't mention the source and destination locations.

But if it's not needed, we can use `-N` feature as you suggested

PR #990, opened 2025-03-05
Update EnsemblSearchDumps_conf.pm

vinay-ebi:
Remove Hive Version 2.5 from pipeline config


nwillhoft:
@vinay-ebi confirmed that patch no longer needed.

PR #993, opened 2025-03-20
Add new script to copy views from one MySQL DB to another

jgtate:
## Description

Adds a standalone bash script that generates shell commands to copy MySQL views from one instance to another. 

## Use case

We make frequent copies of the ensembl_genome_metadata database, but dbcopy doesn't copy views along with the tables. This script will find the views for a given database in a given MySQL instance and generate SQL commands to create those same views in a different instance and database. The SQL commands are wrapped in shell commands that use the MySQL aliases to facilitate connections to the source and target databases.

## Benefits

Hopefully simplifies the process of copying views to new databases.

## Possible Drawbacks

The script doesn't try to actually run the shell commands, it just prints them out. This is because if the views have interdependencies then creating a view could fail unless its dependencies are already available. In that case the CREATE VIEW commands may need to be re-ordered, and it's easier (and safer) to get a human to work that out than to try to code it!

## Testing

Tested by running the generated commands to copy views from mysql-ens-production-1:ensembl_genome_metadata to mysql-ens-test-1:ensembl_genome_metadata_integrated_test.

## Dependencies

The shell where the script is run should have the MySQL aliases available, because it relies on the aliases to simplify connecting to the database instances.

### DIFF
--- a/scripts/copy_views.sh
+++ b/scripts/copy_views.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+# Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+# Copyright [2016-2024] EMBL-European Bioinformatics Institute
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


dpopleton (approved):

nwillhoft:
Hi @jgtate, please can the copyright year on this be updated from 2024 to 2025? It's a minor thing but it's causing the build tests to fail, thanks.

jgtate:
Ah! *That's* what's causing the failures. Sure. I'll sort it out.

PR #998, opened 2025-03-31
Fix file dumps

vinay-ebi:
Fix Hive GUI Load Issue For FTPDUMP Pipeline 
hive db : ens2020_ftp_dumps_beta6_test_gui_1 in h1


dpopleton (approved):
As long as it works for you.

PR #1005, opened 2025-06-05
DumpCore_conf resource class optimization

dpopleton:
Stopped the throttling of the pipeline from a few hive capacity values set to 10.
Replaced these with analysis capacity in these cases.
Removed the batch processing as there is an intrinsic bug to it in hive.
Removed the non-slurm resource classes.

Not tested, but should be fine.


nwillhoft (approved):
Changes look good to optimise pipeline resources.

PR #1006, opened 2025-06-06
Remove support for non-indexed vep folder

likhitha-surapaneni:
## Description

From `release/114`, non-indexed vep cache directory is not supported via FTP dumps. This PR implements it through the VEP file dumping pipeline. It affects variation dumps for vertebrates and non-vertebrates.


nakib103 (approved):
Thanks @likhitha-surapaneni, LGTM!~

vinay-ebi (approved):

PR #1018, opened 2025-08-14
Replace csv with variation handover file & Update track source and de…

Mehrnaz-Charkhchi:
## Description

**Replace CSV files with JSON templates for variation data, Provided by the variation team.
**Update track source handling to prevent adding incomplete sources (missing name or URL).
**Update track descriptions.

## Use case
**Improve variation data loading by migrating from CSV to JSON templates.
**Prevent submission failures caused by missing source urls in db track records.

## Testing

Dependencies
------------

No dependencies!

### DIFF
--- a/nextflow/nf-py-scripts/track_api_loading.py
+++ b/nextflow/nf-py-scripts/track_api_loading.py
@@ -1,6 +1,7 @@
 import argparse
 import csv
 import sys
+import json
 from pathlib import Path
 import yaml
 
@@ -85,7 +86,9 @@ class TrackUtils:
         track_api_json["genome_id"] = self.genome_uuid
         if track_api_json["type"] == "gene":
             if self.custom_gene_track_data:
+                description = track_api_json["description"]+self.custom_gene_track_data[self.genome_uuid]["description_postfix"]
                 track_api_json.update(self.custom_gene_track_data[self.genome_uuid])
+                track_api_json["description"] = description
         if track_api_json["type"] == "variant":
             track_api_json.update(self.custom_variant_track_data)
         return track_api_json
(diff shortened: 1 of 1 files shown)


nextflow/nf-py-scripts/track_api_loading.py
```diff
@@ -193,20 +173,22 @@ def get_gene_track_data(self) -> dict:
                         genome = genome_dataset.genome
```
veidenberg, line 177: I think these lines (170-174 in the updated script) can be removed.

nextflow/nf-py-scripts/track_api_loading.py
```diff
@@ -193,20 +173,22 @@ def get_gene_track_data(self) -> dict:
...
                     for attrib in dataset.dataset_attributes:
                         if attrib.attribute.name == "genebuild.provider_url":
                             source["url"] = attrib.value
                             #row_data.setdefault("sources", [{"url": attrib.value}])
                         if attrib.attribute.name == "genebuild.provider_name":
                             source["name"] = attrib.value
-                        if source:
+                        if source["name"] and source["url"]:
                             row_data.setdefault("sources", [source])
                         if attrib.attribute.name in ARGS.dataset_attributes:
```
veidenberg, line 189: Why is this (`setdefault` for `description_postfix`) needed?

nextflow/nf-py-scripts/track_api_loading.py
```diff
@@ -193,20 +173,22 @@ def get_gene_track_data(self) -> dict:
...
-                            row_data.setdefault("description", "Annotated")
-                    # Set annotation type
-                    if row_data.get("genebuild.provider_name") == "Ensembl":
-                        row_data["description"] = "Imported"
+                            row_data.setdefault("description_postfix", " Genes annotated by Ensembl.")
+                        # Set annotation typei
```
veidenberg, line 193: small typo

nextflow/nf-py-scripts/track_api_loading.py
```diff
@@ -193,20 +173,22 @@ def get_gene_track_data(self) -> dict:
...
+                        if attrib.attribute.name == "genebuild.provider_name" and attrib.value == "Ensembl":
```
veidenberg: This should work, but for some reason the [original DB fetching script](https://github.com/Ensembl/ensembl-web-track-api/blob/9860797f93ec4284a9d442d8ddc7fd24ae4c1395/utils/get_gene_track_desc.py#L94) used `genebuild.annotation_source` instead of `genebuild.provider_name` for choosing the description string (annotated/imported). Better double-check with Stefano and/or datateam.

veidenberg:
It might be a good idea to check the track payload against its datamodel (available [here](https://github.com/Ensembl/ensembl-web-track-api/blob/9860797f93ec4284a9d442d8ddc7fd24ae4c1395/utils/submit_tracks.py#L22)) to catch some errors before submitting (especially since we're now merging external JSON from variation).

veidenberg (approved):
Looks good. Added few sugggestions but it should work as is.

nextflow/nf-py-scripts/track_api_loading.py
Mehrnaz-Charkhchi: Checked with Stefano, it seems correct.

PR #1020, opened 2025-09-12
Switched to pyproject.toml. Simplified requirments

dpopleton:
## Description

Moved the dependencies out of requirments.txt to pyproject.yaml. 
Simplifed them so that we don't specifiy exact version for the entire dependency tree.

## Benefits

Allows modenv to actually be created. 
Takes user work out of dependency installation
Easier for other teams to use.
Less maintence required

## Possible Drawbacks

Could break deployments with minor specifications

Dependencies
------------
Major update to dependencies. This simplifies versioning and prevents conflicts


nwillhoft (approved):
Changes look good to bring the pyproject.toml file in. This will allow dependencies to be categorised to suit different users.

PR #1021, opened 2025-09-15
Feature/embl

mira13:
EMBL format added to file dump pipeline.
Pipeline overall still in progress, so some hardcoded parts or incomplete parts are normal.


dpopleton:
Not sure why the branch is broken wh perl, but you should try to fix that.

dpopleton (approved):
LGTM

PR #1030, opened 2025-11-28
Relax data validation for alphafold data

EbiArnie:
The file containing the alphafold data, accession_ids.csv, now has uniprot accessions with a dash. Adapt code that validates this.


dpopleton (dismissed):

dpopleton:
@EbiArnie . This should be against release 116.

vinay-ebi (approved):

PR #1031, opened 2025-12-11
Update pyproject.toml

dpopleton:
Updated ensembl-py version to fix xrefs


nwillhoft (approved):
Updates to ensembl-py version look good.

PR #1033, opened 2025-12-15
Feature/nf improvements

dpopleton:
We are using this in prod. So it probably should be merged.


vinay-ebi:
It's not production-ready. We can close this PR for now

PR #1042, opened 2026-03-10
Add transcript attrib to JSON output 

vinay-ebi:
Extend genes JSON transcripts to include an attrib object 
Update genes_schema.json to require attrib (with is_canonical) and allow MANE_Select, MANE_Plus_Clinical, apprisq, TSL.


nwillhoft:
Would it also be appropriate to add in GENCODE Primary and Basic transcript attribute info in here too?

We've also added in two new transcript attribute types for 116 (`ens_canon_extended` in https://github.com/Ensembl/ensembl/pull/738 and `readthrough_tra` in https://github.com/Ensembl/ensembl/pull/743), in case they might be relevant here too.

These are transcript attributes that have already been approved for addition into FTP file dumps (and other relevant locations).

dpopleton (approved):

